PR #6, opened 2021-10-12
Dev/spl

alexjhawk:


alexjhawk:
I double-checked the open-source Java String class, and I was wrong in thinking I had adapted our method from that. Thus, I've simply performed an in-place modification of the existing split() method since it would not cause any divergence from a parent source.

PR #16, opened 2022-02-03
Dev/1.6.0

it-hms:
check for file circularized error and exception 
Historical fifo max get behind time enforced 
historical fifo max get behind is settable by library user
moved from file based history EBD to stream based

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Ewon Flexy Extensions Library Changelog
 
+## Version 1.6.0
+### Features
+- Added new method setQueueFifoTimeSpanMins
+- Add maximum duration that historical fifo can get behind (48 hours)
+- Make maximum duration configurable by library user 
+- Changed historical log read to stream from file based 
+- Add check and exception fo circularized file errors
+- Add function for library user to advance historical tracking time
+- Add eventfile module for reading the Flexy event file
+- Add alarm configuration to tag information objects
+### Bug Fixes
+- Fix bug where historical data log overwritten by is circularized file
+- Add missing Javadocs parameters
+
 ## Version 1.5.0
 ### Features
 - Add milliseconds to days, hours, minutes, seconds utility method
--- a/src/main/java/com/hms_networks/americas/sc/extensions/config/ConfigFileAccessManager.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/config/ConfigFileAccessManager.java
@@ -4,7 +4,6 @@ import com.hms_networks.americas.sc.extensions.fileutils.FileAccessManager;
 import com.hms_networks.americas.sc.extensions.json.JSONException;
 import com.hms_networks.americas.sc.extensions.json.JSONObject;
 import com.hms_networks.americas.sc.extensions.json.JSONTokener;
-
 import java.io.IOException;
 
 /**
--- a/src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
@@ -1,0 +1,119 @@
+package com.hms_networks.americas.sc.extensions.eventfile;
+
+import com.ewon.ewonitf.Exporter;
+import com.hms_networks.americas.sc.extensions.historicaldata.EbdTimeoutException;
+import com.hms_networks.americas.sc.extensions.historicaldata.HistoricalDataManager;
+import com.hms_networks.americas.sc.extensions.json.JSONException;
+import com.hms_networks.americas.sc.extensions.string.QuoteSafeStringTokenizer;
+import com.hms_networks.americas.sc.extensions.string.StringUtils;
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Class to manage retrieving events using export block descriptors.
+ *
+ * @author HMS Networks, MU Americas Solution Center
+ * @since 1.6.0
+ */
+public class EventFile {
+  /**
+   * Read EventFile, check for File Circularized event
+   *
+   * @return boolean - did a File Circularized event occur in the recent past
+   * @throws EbdTimeoutException for timeout in EBD read
+   * @throws IOException for errors in parsing response
+   * @throws JSONException for error parsing JSON
+   */
+  public static boolean didFileCircularizedEventOccur()
+      throws IOException, JSONException, EbdTimeoutException {
+    return eventDidOccurInDurationMins(
+        EventFileConstants.DEFAULT_DURATION_MINS, EventFileConstants.FILE_CIRCULARIZED_EVENT_ID);
+  }
+
+  /**
+   * Read EventFile, check for event
+   *
+   * @param eventId event id to look for
+   * @return boolean - did the event occur in the time frame
+   * @throws EbdTimeoutException for timeout in EBD read
+   * @throws IOException for errors in parsing response
+   * @throws JSONException for error parsing JSON
+   */
+  public static boolean eventDidOccur(int eventId)
+      throws IOException, EbdTimeoutException, JSONException {
+    return eventDidOccurInDurationMins(EventFileConstants.DEFAULT_DURATION_MINS, eventId);
+  }
+
+  /**
+   * Read EventFile, check for event
+   *
+   * @param mins duration in minuets to check for event id
+   * @param eventId event id to look for
+   * @return boolean - did the event occur in the time frame
+   * @throws IOException for errors in parsing response
+   * @throws JSONException for error parsing JSON
+   * @throws EbdTimeoutException for Export Block Descriptor timeouts
+   */
+  public static boolean eventDidOccurInDurationMins(int mins, int eventId)
+      throws IOException, JSONException, EbdTimeoutException {
+
+    if (mins < EventFileConstants.MIN_DURATION_MINS
+        || mins > EventFileConstants.MAX_DURATION_MINS) {
+      throw new IllegalArgumentException(
+          "The parameter passed must be greater than "
+              + EventFileConstants.MIN_DURATION_MINS
+              + ", but less than "
+              + EventFileConstants.MAX_DURATION_MINS
+              + ".");
+    }
+
+    String ebdStr = "$dtEV$ftT$st_m" + Integer.toString(mins);
+    Exporter ex = HistoricalDataManager.executeEbdCall(ebdStr);
+    return parseEventFileExportResponse(ex, Long.toString(eventId));
+  }
+
+  /**
+   * Parses Export Block Descriptor Event Log response, checks for event
+   *
+   * @param exporter EBD Exporter
+   * @param eventId ID of the target event as a string
+   * @return was event ID found
+   * @throws IOException for parsing exceptions
+   * @throws JSONException for JSON parsing exceptions
+   */
+  private static boolean parseEventFileExportResponse(Exporter exporter, String eventId)
+      throws IOException, JSONException {
+    final String exporterFile = StringUtils.getStringFromInputStream(exporter, "UTF-8");
+    final List eventFileLines = StringUtils.split(exporterFile, "\n");
+
+    exporter.close();
+    for (int x = 1; x < eventFileLines.size(); x++) {
+      String line = (String) eventFileLines.get(x);
+      if (checkEventFileLine(line.trim(), eventId) == true) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Parse the specified Event file line, return true if id was found.
+   *
+   * @param line line to parse
+   * @param target to check
+   * @return boolean was id found
+   */
+  private static boolean checkEventFileLine(String line, String target) {
+    // Create tokenizer to process line
+    final boolean returnDelimiters = false;
+    QuoteSafeStringTokenizer tokenizer = new QuoteSafeStringTokenizer(line, ";", returnDelimiters);
+
+    // Loop through each token
+    String currentToken = "";
+    while (tokenizer.hasMoreElements()) {
+      // Get next token
+      currentToken = tokenizer.nextToken();
+    }
+    return currentToken.equals(target);
+  }
+}
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/CircularizedFileException.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/CircularizedFileException.java
@@ -1,0 +1,13 @@
+package com.hms_networks.americas.sc.extensions.historicaldata;
+
+/**
+ * Class for Circularized File Exception.
+ *
+ * @author HMS Networks, MU Americas Solution Center
+ * @since 1.6.0
+ */
+public class CircularizedFileException extends Exception {
+  public CircularizedFileException(String errorMessage) {
+    super(errorMessage);
+  }
+}
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
@@ -52,4 +52,7 @@ class HistoricalDataConstants {
 
   /** Name of historical data queue time tracker file 2. */
   static final String QUEUE_TIME_FILE_2_NAME = "histDataTime2";
+
+  /** Minimum amount of Historical FIFO queue that can be set by library users. */
+  static final long MIN_QUEUE_SPAN_MINS = 1;
 }
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
@@ -60,4 +60,10 @@ class HistoricalDataConstants {
 
   /** Default maximum amount of time Historical FIFO can get behind. */
   static final long MAX_HIST_FIFO_GET_BEHIND_MS = SCTimeUnit.HOURS.toMillis(12);
+
+  /** Default thread sleep time waiting for EBD in ms. */
+  static final int DEFAULT_EBD_THREAD_SLEEP_MS = 5;
+
+  /** Maximum amount of time that EBD call should wait in milliseconds before timing out. */
+  static final long MAX_EBD_WAIT_MS = 15000;
 }
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
@@ -1,5 +1,7 @@
 package com.hms_networks.americas.sc.extensions.historicaldata;
 
+import com.hms_networks.americas.sc.extensions.system.time.SCTimeUnit;
+
 /**
  * Class for storing constants used in the queueing package.
  *
@@ -55,4 +57,7 @@ class HistoricalDataConstants {
 
   /** Minimum amount of Historical FIFO queue that can be set by library users. */
   static final long MIN_QUEUE_SPAN_MINS = 1;
+
+  /** Default maximum amount of time Historical FIFO can get behind. */
+  static final long MAX_HIST_FIFO_GET_BEHIND_MS = SCTimeUnit.HOURS.toMillis(12);
 }
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
@@ -9,6 +9,7 @@ import com.hms_networks.americas.sc.extensions.datapoint.DataPointInteger;
 import com.hms_networks.americas.sc.extensions.datapoint.DataPointIntegerMappedString;
 import com.hms_networks.americas.sc.extensions.datapoint.DataPointString;
 import com.hms_networks.americas.sc.extensions.datapoint.DataQuality;
+import com.hms_networks.americas.sc.extensions.eventfile.EventFile;
 import com.hms_networks.americas.sc.extensions.fileutils.FileConstants;
 import com.hms_networks.americas.sc.extensions.json.JSONException;
 import com.hms_networks.americas.sc.extensions.string.QuoteSafeStringTokenizer;
@@ -135,9 +136,11 @@ public class HistoricalDataManager {
    * @return dataPoints
    * @throws IOException for parsing Exceptions
    * @throws JSONException for JSON parsing Exceptions
+   * @throws EbdTimeoutException timeout for reading event file
+   * @throws CircularizedFileException when event log indicates circularized event
    */
   private static ArrayList parseEBDHistoricalLogExportResponse(Exporter exporter)
-      throws IOException, JSONException {
+      throws IOException, JSONException, EbdTimeoutException, CircularizedFileException {
 
     final String exporterFile = StringUtils.getStringFromInputStream(exporter, "UTF-8");
     final List eventFileLines = StringUtils.split(exporterFile, "\n");
@@ -151,6 +154,11 @@ public class HistoricalDataManager {
     }
 
     exporter.close();
+
+    // Check for Circularized Event
+    if (EventFile.didFileCircularizedEventOccur()) {
+      throw new CircularizedFileException("A circularized event was found in the event logs.");
+    }
     return dataPoints;
   }
 
@@ -201,6 +209,7 @@ public class HistoricalDataManager {
    * @throws IOException if export block descriptor fails
    * @throws JSONException if unable to parse int to string enumeration file
    * @throws EbdTimeoutException for EBD timeout
+   * @throws CircularizedFileException if circularized file event is found
    */
   public static ArrayList readHistoricalFifo(
       String startTime,
@@ -210,7 +219,7 @@ public class HistoricalDataManager {
       boolean includeTagGroupC,
       boolean includeTagGroupD,
       boolean stringHistorical)
-      throws IOException, JSONException, EbdTimeoutException {
+      throws IOException, JSONException, EbdTimeoutException, CircularizedFileException {
 
     // create EBD string
     final String ebdStr =
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
@@ -346,6 +346,8 @@ public class HistoricalDataManager {
       // Read next line before looping again
       line = reader.readLine();
     }
+    reader.close();
+
     return dataPoints;
   }
 
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
@@ -1,7 +1,6 @@
 package com.hms_networks.americas.sc.extensions.historicaldata;
 
 import com.ewon.ewonitf.Exporter;
-
 import com.hms_networks.americas.sc.extensions.datapoint.DataPoint;
 import com.hms_networks.americas.sc.extensions.datapoint.DataPointBoolean;
 import com.hms_networks.americas.sc.extensions.datapoint.DataPointDword;
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
@@ -12,6 +12,7 @@ import com.hms_networks.americas.sc.extensions.datapoint.DataQuality;
 import com.hms_networks.americas.sc.extensions.fileutils.FileConstants;
 import com.hms_networks.americas.sc.extensions.json.JSONException;
 import com.hms_networks.americas.sc.extensions.string.QuoteSafeStringTokenizer;
+import com.hms_networks.americas.sc.extensions.string.StringUtils;
 import com.hms_networks.americas.sc.extensions.taginfo.TagInfo;
 import com.hms_networks.americas.sc.extensions.taginfo.TagInfoEnumeratedIntToString;
 import com.hms_networks.americas.sc.extensions.taginfo.TagInfoManager;
@@ -20,6 +21,7 @@ import java.io.BufferedReader;
 import java.io.FileReader;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Class to manage retrieving tag information and historical logs using export block descriptors.
@@ -55,6 +57,175 @@ public class HistoricalDataManager {
         includeTagGroupC,
         includeTagGroupD,
         stringHistorical);
+  }
+
+  /**
+   * Prepare Read Historical Log Export Block Descriptor (EBD) string.
+   *
+   * @param startTime start time of export
+   * @param endTime end time of export
+   * @param includeTagGroupA include tag group A
+   * @param includeTagGroupB include tag group B
+   * @param includeTagGroupC include tag group C
+   * @param includeTagGroupD include tag group D
+   * @param stringHistorical export string historical logs if true
+   * @return EBD string
+   */
+  private static String prepareHistoricalFifoReadEBDString(
+      String startTime,
+      String endTime,
+      boolean includeTagGroupA,
+      boolean includeTagGroupB,
+      boolean includeTagGroupC,
+      boolean includeTagGroupD,
+      boolean stringHistorical) {
+
+    // Check for valid group selection
+    if (!includeTagGroupA && !includeTagGroupB && !includeTagGroupC && !includeTagGroupD) {
+      throw new IllegalArgumentException(
+          "Cannot generate historical logs with no tag groups selected.");
+    }
+
+    // Build string of tag groups for filter type
+    String tagGroupFilterStr = "";
+    if (includeTagGroupA) {
+      tagGroupFilterStr += "A";
+    }
+    if (includeTagGroupB) {
+      tagGroupFilterStr += "B";
+    }
+    if (includeTagGroupC) {
+      tagGroupFilterStr += "C";
+    }
+    if (includeTagGroupD) {
+      tagGroupFilterStr += "D";
+    }
+
+    // Get EBD data type
+    String ebdDataType;
+    if (stringHistorical) {
+      ebdDataType = "HS";
+    } else {
+      ebdDataType = "HL";
+    }
+
+    /*
+     * Build EBD string with parameters
+     * dt[ebdDataType]: data type, value specified in string ebdDataType
+     * ftT: file type, text
+     * startTime: start time for data
+     * endTime: end time for data
+     * flABCD: filter type, specified tag groups
+     */
+    return "$dt"
+        + ebdDataType
+        + "$ftT$st"
+        + startTime
+        + "$et"
+        + endTime
+        + "$fl"
+        + tagGroupFilterStr;
+  }
+
+  /**
+   * Parses Export Block Descriptor Historical Log response into dataPoints. Note: this function
+   * only handles Historical List responses.
+   *
+   * @param exporter EBD Exporter
+   * @return dataPoints
+   * @throws IOException for parsing Exceptions
+   * @throws JSONException for JSON parsing Exceptions
+   */
+  private static ArrayList parseEBDHistoricalLogExportResponse(Exporter exporter)
+      throws IOException, JSONException {
+
+    final String exporterFile = StringUtils.getStringFromInputStream(exporter, "UTF-8");
+    final List eventFileLines = StringUtils.split(exporterFile, "\n");
+    ArrayList dataPoints = new ArrayList();
+    for (int x = 1; x < eventFileLines.size(); x++) {
+      String line = (String) eventFileLines.get(x);
+      DataPoint lineDataPoint = parseHistoricalFileLine(line.trim());
+      if (lineDataPoint != null) {
+        dataPoints.add(lineDataPoint);
+      }
+    }
+
+    exporter.close();
+    return dataPoints;
+  }
+
+  /**
+   * Executes EBD call, waits for data return.
+   *
+   * @param ebdStr EBD string
+   * @return EBD Exporter - caller must close exporter
+   * @throws EbdTimeoutException when there is no response before timeout period
+   * @throws IOException when there is an Exporter Exception
+   */
+  public static Exporter executeEbdCall(String ebdStr) throws IOException, EbdTimeoutException {
+    final long start = System.currentTimeMillis();
+    final Exporter exporter = new Exporter(ebdStr);
+    int available = exporter.available();
+    // check on data availability
+    while (available == 0) {
+      // sleep
+      try {
+        Thread.sleep(HistoricalDataConstants.DEFAULT_EBD_THREAD_SLEEP_MS);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
+      }
+      // check for timeout
+      if (System.currentTimeMillis() - start > HistoricalDataConstants.MAX_EBD_WAIT_MS) {
+        exporter.close();
+        throw new EbdTimeoutException("Timeout waiting for Export Block Descriptor.");
+      }
+
+      // check for data available
+      available = exporter.available();
+    }
+    return exporter;
+  }
+
+  /**
+   * Reads historical log between <code>startTime</code> and <code>endTime</code>. Returns
+   * Datapoints.
+   *
+   * @param startTime start time of export
+   * @param endTime end time of export
+   * @param includeTagGroupA include tag group A
+   * @param includeTagGroupB include tag group B
+   * @param includeTagGroupC include tag group C
+   * @param includeTagGroupD include tag group D
+   * @param stringHistorical export string historical logs if true
+   * @return data points from response
+   * @throws IOException if export block descriptor fails
+   * @throws JSONException if unable to parse int to string enumeration file
+   * @throws EbdTimeoutException for EBD timeout
+   */
+  public static ArrayList readHistoricalFifo(
+      String startTime,
+      String endTime,
+      boolean includeTagGroupA,
+      boolean includeTagGroupB,
+      boolean includeTagGroupC,
+      boolean includeTagGroupD,
+      boolean stringHistorical)
+      throws IOException, JSONException, EbdTimeoutException {
+
+    // create EBD string
+    final String ebdStr =
+        prepareHistoricalFifoReadEBDString(
+            startTime,
+            endTime,
+            includeTagGroupA,
+            includeTagGroupB,
+            includeTagGroupC,
+            includeTagGroupD,
+            stringHistorical);
+
+    // Execute EBD call and parse results
+    final Exporter exporter = executeEbdCall(ebdStr);
+    return parseEBDHistoricalLogExportResponse(exporter);
   }
 
   /**
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
@@ -178,11 +178,12 @@ public class HistoricalDataQueueManager {
    * @throws TimeTrackerUnrecoverableException if both time tracking files are corrupted
    * @throws CorruptedTimeTrackerException if the current time tracking file is corrupted
    * @throws JSONException if unable to parse int to string enumeration file
-   * @throws EbdTimeoutException for EBD timeout
+   * @throws CircularizedFileException if circularized file exception was found
+   * @throws EbdTimeoutException for EBD timeouts
    */
   public static synchronized ArrayList getFifoNextSpanDataAllGroups(boolean startNewTimeTracker)
       throws IOException, TimeTrackerUnrecoverableException, CorruptedTimeTrackerException,
-          JSONException, EbdTimeoutException {
+          JSONException, CircularizedFileException, EbdTimeoutException {
     final boolean includeTagGroupA = true;
     final boolean includeTagGroupB = true;
     final boolean includeTagGroupC = true;
@@ -311,6 +312,7 @@ public class HistoricalDataQueueManager {
    * @throws CorruptedTimeTrackerException one of the tracking files is corrupted
    * @throws JSONException if unable to parse int to string enumeration file
    * @throws EbdTimeoutException when EBD call times out
+   * @throws CircularizedFileException if circularized file exception was found
    */
   public static synchronized ArrayList getFifoNextSpanData(
       boolean startNewTimeTracker,
@@ -319,7 +321,7 @@ public class HistoricalDataQueueManager {
       boolean includeTagGroupC,
       boolean includeTagGroupD)
       throws IOException, TimeTrackerUnrecoverableException, CorruptedTimeTrackerException,
-          JSONException, EbdTimeoutException {
+          JSONException, EbdTimeoutException, CircularizedFileException {
 
     if (!hasInitTime) {
       initTimeTrackerFiles();
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
@@ -71,8 +71,15 @@ public class HistoricalDataQueueManager {
    * Set the FIFO queue time span in minutes.
    *
    * @param timeSpanMins new FIFO queue time span in minutes
+   * @throws IllegalArgumentException if unable to read a file
    */
   public static synchronized void setQueueFifoTimeSpanMins(long timeSpanMins) {
+    if (timeSpanMins < HistoricalDataConstants.MIN_QUEUE_SPAN_MINS) {
+      throw new IllegalArgumentException(
+          "The FifoTimeSpan must not be less than "
+              + HistoricalDataConstants.MIN_QUEUE_SPAN_MINS
+              + ".");
+    }
     queueFifoTimeSpanMins = timeSpanMins;
   }
 
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
@@ -327,87 +327,60 @@ public class HistoricalDataQueueManager {
 
     ArrayList queueData = new ArrayList();
 
-    // Ensure start time is not equal to end time, and enough time has passed to grab new data
-    if (enoughSeconds(startTimeTrackerMsLong, endTimeTrackerMsLong)) {
-      // Calculate EBD start and end time
-      final String ebdStartTime = convertToEBDTimeFormat(startTimeTrackerMsLong);
-      final String ebdEndTime = convertToEBDTimeFormat(endTimeTrackerMsLong);
+    // Calculate EBD start and end time
+    final String ebdStartTime = convertToEBDTimeFormat(startTimeTrackerMsLong);
+    final String ebdEndTime = convertToEBDTimeFormat(endTimeTrackerMsLong);
 
-      // Run standard EBD export call (int, float, ...)
-      final String ebdFileName =
+    // Run standard EBD export call (int, float, ...)
+    final String ebdFileName =
+        HistoricalDataConstants.QUEUE_FILE_FOLDER
+            + "/"
+            + HistoricalDataConstants.QUEUE_EBD_FILE_NAME
+            + HistoricalDataConstants.QUEUE_FILE_EXTENSION;
+    boolean stringHistorical = false;
+    HistoricalDataManager.exportHistoricalToFile(
+        ebdStartTime,
+        ebdEndTime,
+        ebdFileName,
+        includeTagGroupA,
+        includeTagGroupB,
+        includeTagGroupC,
+        includeTagGroupD,
+        stringHistorical);
+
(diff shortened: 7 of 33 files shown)


it-hms:
@TomKimsey The first PR was closed so I could add Alex as reviewer.
@alexjhawk  I did force push without realizing you had added commits. Can you cherry-pick those commits?

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Ewon Flexy Extensions Library Changelog
 
+## Version 1.6.0
```
TomKimsey, line 3: I assume this is the point at which you would tag the repo for a release. I dont think its documented but we have been setting the commit title to "RELEASE vx.y.z" to consistently mark this. 


@alexjhawk  should we add this to CONTRIBUTING.md?

alexjhawk:
> @TomKimsey The first PR was closed so I could add Alex as reviewer. @alexjhawk I did force push without realizing you had added commits. Can you cherry-pick those commits? Are those commits on a different branch that I can cherry pick.

I have a backup of the files locally, but everything was on the dev/1.6.0 branch, so it has been lost on the GitHub side. Additionally, there should likely also be cleanup performed on the Maven repository now. I had pushed multiple pre-release versions for testing during development which are no longer applicable.

CHANGELOG.md
alexjhawk: @TomKimsey  Yes, this should be added. Incrementing the version and any associated change logs needs to be in the last commit. Here is an example: https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/commit/294353ed050dfb7bc6dc395d4faf91a1ed619555.

The formatting in CHANGELOG.md is especially important, as it could cause the build and release actions to fail if not followed.

TomKimsey:
> @TomKimsey The first PR was closed so I could add Alex as reviewer. @alexjhawk I did force push without realizing you had added commits. Can you cherry-pick those commits? Are those commits on a different branch that I can cherry pick.

@it-hms --force-with-lease is a safer option than -f/--force. If someone else has added commits it should reject your push.

it-hms:
@TomKimsey  good call on the --force-with-lease

CHANGELOG.md
it-hms: ok.. will change commit title.

it-hms:
@alexjhawk @TomKimsey  I have cherry picked Alex's commit. It's read for review (assuming the checks pass)

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
```diff
@@ -0,0 +1,130 @@
...
+   * @throws IOException for errors in parsing response
+   * @throws JSONException for error parsing JSON
+   */
+  public static boolean EventDidOccurInDurationMins(int mins, int eventId)
+      throws IOException, TimeoutException, JSONException {
+
+    if (mins < EventFileConstants.MIN_DURATION_MINS
+        || mins > EventFileConstants.MAX_DURATION_MINS) {
+      throw new IllegalArgumentException(
+          "The parameter passed must be greater than 0, but less than 90.");
```
TomKimsey, line 64:
```suggestion
          "The parameter passed must be greater than " + EventFileConstants.MIN_DURATION_MINS + ", but less than " + EventFileConstants.MAX_DURATION_MINS + " .");
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -52,4 +52,16 @@
 
   /** Name of historical data queue time tracker file 2. */
   static final String QUEUE_TIME_FILE_2_NAME = "histDataTime2";
+
+  /** Minimum amount of Historical FIFO queue that can be set by library users. */
+  static final long MIN_QUEUE_SPAN_MINS = 1;
+
+  /** Default maximum amount of time Historical FIFO can get behind. 48hours in ms. */
```
TomKimsey, line 59:
```suggestion
  /** Default maximum amount of time Historical FIFO can get behind. 48 hours in ms. */
```

or
```suggestion
  /** Default maximum amount of time Historical FIFO can get behind. */
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -52,4 +52,16 @@
...
+  static final long MAX_HIST_FIFO_GET_BEHIND_MS = 4147200000L;
+
+  /** Default thread sleep time waiting for EBD in ms. */
+  static final int THREAD_SLEEP_MS = 5;
```
TomKimsey, line 63:
```suggestion
  static final int DEFAULT_EBD_THREAD_SLEEP_MS = 5;
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -58,6 +61,200 @@ public static void exportAllHistoricalToFile(
         stringHistorical);
   }
 
+  /**
+   * Prepare Read Historical Log Export Bloc Descriptor (EBD) string.
```
TomKimsey, line 65:
```suggestion
   * Prepare Read Historical Log Export Block Descriptor (EBD) string.
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -58,6 +61,200 @@ public static void exportAllHistoricalToFile(
...
+        + startTime
+        + "$et"
+        + endTime
+        + "$fl"
+        + tagGroupFilterStr;
+  }
+
+  /**
+   * Parses Export Block Descriptor Historical Log response into dataPoints. Note: this function
+   * only handles Historical List responses
```
TomKimsey, line 134:
```suggestion
   * only handles Historical List responses.
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -58,6 +61,200 @@ public static void exportAllHistoricalToFile(
...
+
+  /**
+   * Executes EBD call, waits for data return.
+   *
+   * @param ebdStr EBD string
+   * @return EBD Exporter - caller must close exporter
+   * @throws TimeoutException when there is no response before timeout period
+   * @throws IOException when there is an Exporter Exception
+   */
+  public static Exporter ExecuteEBDCall(String ebdStr) throws IOException, TimeoutException {
```
TomKimsey, line 188:
```suggestion
  public static Exporter ExecuteEbdCall(String ebdStr) throws IOException, TimeoutException {
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -58,6 +61,200 @@ public static void exportAllHistoricalToFile(
...
+      // sleep
+      try {
+        Thread.sleep(HistoricalDataConstants.THREAD_SLEEP_MS);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
+      }
+      // check for timeout
+      if (System.currentTimeMillis() - start > HistoricalDataConstants.MAX_EBD_WAIT) {
+        exporter.close();
+        throw new TimeoutException("could not wait any longer for EBD call");
```
TomKimsey, line 203:
```suggestion
        throw new TimeoutException("Could not wait any longer for EBD call.");
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -58,6 +61,200 @@ public static void exportAllHistoricalToFile(
...
+            includeTagGroupC,
+            includeTagGroupD,
+            stringHistorical);
+
+    // Execute EBD call and parse results
+    try {
+      final Exporter exporter = ExecuteEBDCall(ebdStr);
+      return parseEBDHistoricalLogExportResponse(exporter);
+    } catch (TimeoutException e) {
+      return new ArrayList();
```
TomKimsey, line 254: No logging of the timeout?

src/main/java/com/hms_networks/americas/sc/extensions/config/ConfigFileAccessManager.java
```diff
@@ -4,7 +4,6 @@
 import com.hms_networks.americas.sc.extensions.json.JSONException;
 import com.hms_networks.americas.sc.extensions.json.JSONObject;
 import com.hms_networks.americas.sc.extensions.json.JSONTokener;
-
 import java.io.IOException;
```
TomKimsey, line 7: unrelated to javadoc errors commit 76cc5df25326306a

CHANGELOG.md
TomKimsey: Commit title should be "RELEASE: v1.6.0"

TomKimsey (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -70,12 +72,42 @@ public static synchronized long getQueueFifoTimeSpanMins() {
...
 
+  /**
+   * Set the FIFO queue time span in minutes.
+   *
+   * @param timeSpanMins new FIFO queue time span in minutes
+   * @throws IllegalArgumentException if unable to read a file
+   */
+  public static synchronized void setQueueFifoTimeSpanMinsCheck(long timeSpanMins) {
+    if (timeSpanMins < HistoricalDataConstants.MIN_QUEUE_SPAN_MINS) {
+      throw new IllegalArgumentException("The FifoTimeSpan must not be less than one.");
```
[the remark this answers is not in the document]
alexjhawk: I think this should just be added to the existing #setQueueFifoTimeSpanMins() method. The old method shouldn't need to be deprecated, because the behavior you're introducing is just a check of an existing constraint. Nothing should be affected by the change unless it was already misusing the library, so this would require unnecessary changes in existing connectors.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
TomKimsey: I generally like to leave the value out of comments. If someone changes the value there is a chance they may not update the comment, then its a code maintenance issue. If someone really wanted to know what the value was they could just use a calculator.
image.

Another way to do it and have it documented in the code is 
`
long MAX_HIST_FIFO_GET_BEHIND_HOURS = 48L;
long MAX_HIST_FIFO_GET_BEHIND_MS = MAX_HIST_FIFO_GET_BEHIND_HOURS * HOURS_PER_DAY * MINUTES_PER_HOUR * SECONDS_PER_MINUTE * MILLIS_PER_SECOND;
`

But thats a bit nasty looking

OR cleaner would be 
`
long MAX_HIST_FIFO_GET_BEHIND_HOURS = 48L;
long MAX_HIST_FIFO_GET_BEHIND_MS = MAX_HIST_FIFO_GET_BEHIND_HOURS * MILLISECONDS_PER_DAY;
`

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -334,87 +334,61 @@ public static synchronized ArrayList getFifoNextSpanData(
...
+    final String newTimeTrackerVal = Long.toString(endTimeTrackerMsLong + 1);
+    FileAccessManager.writeStringToFile(writeFile, newTimeTrackerVal);
+
     isFile1CurrTimeTrackerFile = !isFile1CurrTimeTrackerFile;
 
     // Return data
     return queueData;
   }
 
-  /**
```
alexjhawk, line 384: Has this been tested thoroughly? I cannot remember the details, but I thought that enoughSeconds() was implemented due to a bug we saw, and it fixed that behavior. I think Oliver has taken that on. See https://github.com/hms-networks/sc-flexy-tag-data-lib/commit/ce6fe4e11e44eea67f725a86ba00d893685314bf

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
TomKimsey: wait, somethings not right here. let me look at this

TomKimsey: @it-hms I think you have 48 days in milliseconds here not 48 hours.

alexjhawk: This can be simplified further using our TimeUnit class.

For example,

```java
    int millisInOneHour = SCTimeUnit.HOURS.toMillis(1);
    int millisInFourHours = SCTimeUnit.HOURS.toMillis(4);
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -70,12 +72,42 @@ public static synchronized long getQueueFifoTimeSpanMins() {
...
+  }
+
+  /**
+   * Set the maximum amount the Historical FIFO queue can get behind in minutes.
+   *
+   * @param timeMins new FIFO queue time span in minutes
+   * @throws IllegalArgumentException if parameter is negative
+   */
+  public static synchronized void setQueueMaxBehindMins(long timeMins) {
+    if (timeMins < 0) {
```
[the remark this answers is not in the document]
it-hms: I did think that someone might want to set this to zero, but on second consideration I guess that doesn't make too much sense. They would never receive any data.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -89,6 +92,22 @@ public static synchronized void setQueueFifoTimeSpanMinsCheck(long timeSpanMins)
...
+      throw new IllegalArgumentException("The parameter passed must be greater than 0.");
+    }
+    maxQueueGetsBehindMs =
+        timeMins
+            * HistoricalDataConstants.TIME_SECS_PER_MIN
+            * HistoricalDataConstants.TIME_MS_PER_SEC;
```
alexjhawk, line 108:
```suggestion
    maxQueueGetsBehindMs = SCTimeUnit.MINUTES.toMilliseconds(timeMins);
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
alexjhawk: I don't think the proper way to handle a timeout is in this library. It prevents the connector from having visibility into the timeout and instead it looks like it just gets a list of no datapoints. Shouldn't the connector that implements it be able to catch the TimeoutException and either log it or retry?

src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
```diff
@@ -561,7 +561,7 @@ public int length() {
     return ja.length() == 0 ? null : ja;
   }
 
-  /** Shave off trailing zeros and decimal point, if possible. */
+  /* Shave off trailing zeros and decimal point, if possible. */
```
alexjhawk, line 564: The method is now missing any Javadoc with this change.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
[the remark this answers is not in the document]
it-hms: resolved

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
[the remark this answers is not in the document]
TomKimsey: This is based on your comment in the message 

`throw new IllegalArgumentException("The parameter passed must be greater than 0.");`

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
alexjhawk: Have you tested this on the Flexy? TimeoutException reports as Java 1.5 or newer: 
image

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -136,9 +137,11 @@ private static String prepareHistoricalFifoReadEBDString(
    * @return dataPoints
    * @throws IOException for parsing Exceptions
    * @throws JSONException for JSON parsing Exceptions
+   * @throws TimeoutException timeout for reading event file
+   * @throws CircularizedFileException when event log indicates circularized event
    */
   private static ArrayList parseEBDHistoricalLogExportResponse(Exporter exporter)
-      throws IOException, JSONException {
+      throws IOException, JSONException, TimeoutException, CircularizedFileException {
```
alexjhawk, line 144: See earlier comment about TimeoutException being documented as Java 1.5 and newer.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -401,13 +387,67 @@ public static synchronized ArrayList getFifoNextSpanData(
...
+
+  /**
+   * Update Historical FIFO tracking time
+   *
+   * @param updateTime the time to be written
+   * @throws IOException if unable to read or write files
+   * @throws TimeTrackerUnrecoverableException if both time tracking files are corrupted
+   * @throws CorruptedTimeTrackerException one of the tracking files is corrupted
+   */
+  private static void UpdateTrackingStartTime(long updateTime)
```
alexjhawk, line 434: Method name does not follow the required format. It should be camel case, so

updateTrackingStartTime()

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -450,4 +450,28 @@ private static void UpdateTrackingStartTime(long updateTime)
...
+
+  /**
+   * Advance Historical FIFO tracking time. The intention is for exception handlers to make the
+   * decision to advance time.
+   *
+   * @throws IOException if unable to read or write files
+   * @throws TimeTrackerUnrecoverableException if both time tracking files are corrupted
+   * @throws CorruptedTimeTrackerException one of the tracking files is corrupted
+   */
+  public static void AdvanceTrackingStartTime()
```
alexjhawk, line 462: Should be advanceTrackingStartTime() (camel case)

CHANGELOG.md
alexjhawk: +1

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
```diff
@@ -0,0 +1,130 @@
...
+    // Create line counter and variable to store current line
+    int lineCount = 0;
+    String line = reader.readLine();
+    // Loop through lines in file until end and store data points
+    while (line != null) {
+
+      // Only parse lines 1 and greater, skip header
+      if (lineCount > 0) {
+        // Parse line
+        if (checkEventFileLine(line, eventId) == true) {
```
alexjhawk, line 95: Redundant if statement can be simplified to 
```java
          return checkEventFileLine(line, eventId);
```

src/main/java/com/hms_networks/americas/sc/extensions/config/ConfigFileAccessManager.java
it-hms: resolved... will put in separate commit for formatting

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
```diff
@@ -0,0 +1,130 @@
...
+      }
+
+      // Increment line count
+      lineCount++;
+      // Read next line before looping again
+      line = reader.readLine();
+    }
+    reader.close();
+    exporter.close();
+    return false;
```
alexjhawk, line 107:
```suggestion
    final String exporterFile = StringUtils.getStringFromInputStream(exporter, "UTF-8");
    final List eventFileLines = StringUtils.split(exporterFile, "\n");
    for (int x = 0; x < eventFileLines.size(); x++) {
      String eventFileLine = (String) eventFileLines.get(x);
      return checkEventFileLine(eventFileLine, eventId);
    }

    exporter.close();
    return false;
```

I think it would be better to use the StringUtils function for reading the exporter. It allows for an overall simpler and smaller method.

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -58,6 +61,200 @@ public static void exportAllHistoricalToFile(
...
+   * @param exporter EBD Exporter
+   * @return dataPoints
+   * @throws IOException for parsing Exceptions
+   * @throws JSONException for JSON parsing Exceptions
+   * @throws TimeoutException timeout for reading event file
+   * @throws CircularizedFileException when event log indicates circularized event
+   */
+  private static ArrayList parseEBDHistoricalLogExportResponse(Exporter exporter)
+      throws IOException, JSONException, TimeoutException, CircularizedFileException {
+    final DataInputStream dataStream = new DataInputStream(exporter);
```
alexjhawk, line 145: I think this should also follow the simpler design offered in my comment on the event file manager https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/pull/16#discussion_r798914601

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
alexjhawk: Shouldn't this be executeEbdCall() instead?

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: Thanks for checking this ... resolved.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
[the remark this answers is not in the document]
it-hms: based on my comment @throws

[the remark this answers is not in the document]
it-hms: resolved

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: ok, changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: ok. Changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
alexjhawk: Yes, I couldn't remember if it was millis or milliseconds

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: changed

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: 👍🏼  :(  you are correct. Why can't we have nice things?

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
it-hms: I have manually tested this with no abnormal behavior.  It always returns just the header.  I did abuse it a bit too ( tried the 33rd day of January, tried the 75th second) and it stood up... it just returns the header.  
How much you want to be about there being a bug on February 29th of a leap year?

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: changed

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: Changed to custom timeout class.

src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
it-hms: I'll bring back the docs. Changed.

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
it-hms: resolved

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
it-hms: This is in a while loop. I don't need it to return the first result. I need it to return the first true it finds or false if there are none.

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
it-hms: I care more about memory efficiency. StringUtils copies the bytes into a new string. Then copies parts of that string in to a new arrayList. This means that every byte from the stream is copied twice.  With all the Flexy memory and thread starvation uncertainty, my preference is for memory efficiency and performance. My hope is that bufferedReader is somehow more efficient with memory.  I really am not familiar with Java so I will make the change if you feel it is the better choice.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: changed to EBdTimeout custom class

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
it-hms: updated.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
it-hms: changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: I left a comment about this someplace else, but my main concern is memory efficiency. Normally simpler is more efficient, but I know that StringUtils makes additional copies. I hope that buffered reader is more efficient.

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
alexjhawk: I designed the StringUtils.getStringFromInputStream method with performance in mind. I did research online and generally noticed that ByteArrayOutputStream/InputStream.read was the most efficient out of most solutions. Additionally, BufferedReader can/will modify the contents to convert any line separators to the system line separator, which we may not want coming from a stream, where we expect consistent output.

See: https://stackoverflow.com/questions/309424/how-do-i-read-convert-an-inputstream-into-a-string-in-java

alexjhawk: Additionally, if there is a memory or performance concern with something in the library, I think we should focus on resolving it in the shared method instead of duplicating code across the library that does the same thing(s).

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
alexjhawk: It may be the \n character or some other control character in the ASCII table.

alexjhawk: I replied to your other comment about memory efficiency/performance and included a link that references performance:

--------------------------------

Performance tests

Performance tests for small String (length = 175), url in [github](https://github.com/Vedenin/useful-java-links/blob/master/helloworlds/5.0-other-examples/src/main/java/other_examples/ConvertInputStreamToStringBenchmark.java) (mode = Average Time, system = Linux, score 1,343 is the best):

```
              Benchmark                         Mode  Cnt   Score   Error  Units
 8. ByteArrayOutputStream and read (JDK)        avgt   10   1,343 ± 0,028  us/op
 6. InputStreamReader and StringBuilder (JDK)   avgt   10   6,980 ± 0,404  us/op
10. BufferedInputStream, ByteArrayOutputStream  avgt   10   7,437 ± 0,735  us/op
11. InputStream.read() and StringBuilder (JDK)  avgt   10   8,977 ± 0,328  us/op
 7. StringWriter and IOUtils.copy (Apache)      avgt   10  10,613 ± 0,599  us/op
 1. IOUtils.toString (Apache Utils)             avgt   10  10,605 ± 0,527  us/op
 3. Scanner (JDK)                               avgt   10  12,083 ± 0,293  us/op
 2. CharStreams (guava)                         avgt   10  12,999 ± 0,514  us/op
 4. Stream Api (Java 8)                         avgt   10  15,811 ± 0,605  us/op
 9. BufferedReader (JDK)                        avgt   10  16,038 ± 0,711  us/op
 5. parallel Stream Api (Java 8)                avgt   10  21,544 ± 0,583  us/op

```
Performance tests for big String (length = 50100), url in [github](https://github.com/Vedenin/useful-java-links/blob/master/helloworlds/5.0-other-examples/src/main/java/other_examples/ConvertBigStringToInputStreamBenchmark.java)[](https://i.stack.imgur.com/AYYhz.png) (mode = Average Time, system = Linux, score 200,715 is the best):

```
               Benchmark                        Mode  Cnt   Score        Error  Units
 8. ByteArrayOutputStream and read (JDK)        avgt   10   200,715 ±   18,103  us/op
 1. IOUtils.toString (Apache Utils)             avgt   10   300,019 ±    8,751  us/op
 6. InputStreamReader and StringBuilder (JDK)   avgt   10   347,616 ±  130,348  us/op
 7. StringWriter and IOUtils.copy (Apache)      avgt   10   352,791 ±  105,337  us/op
 2. CharStreams (guava)                         avgt   10   420,137 ±   59,877  us/op
 9. BufferedReader (JDK)                        avgt   10   632,028 ±   17,002  us/op
 5. parallel Stream Api (Java 8)                avgt   10   662,999 ±   46,199  us/op
 4. Stream Api (Java 8)                         avgt   10   701,269 ±   82,296  us/op
10. BufferedInputStream, ByteArrayOutputStream  avgt   10   740,837 ±    5,613  us/op
 3. Scanner (JDK)                               avgt   10   751,417 ±   62,026  us/op
11. InputStream.read() and StringBuilder (JDK)  avgt   10  2919,350 ± 1101,942  us/op
```
Graphs (performance tests depending on Input Stream length in Windows 7 system)
image

Performance test (Average Time) depending on Input Stream length in Windows 7 system:

```
 length  182    546     1092    3276    9828    29484   58968

 test8  0.38    0.938   1.868   4.448   13.412  36.459  72.708
 test4  2.362   3.609   5.573   12.769  40.74   81.415  159.864
 test5  3.881   5.075   6.904   14.123  50.258  129.937 166.162
 test9  2.237   3.493   5.422   11.977  45.98   89.336  177.39
 test6  1.261   2.12    4.38    10.698  31.821  86.106  186.636
 test7  1.601   2.391   3.646   8.367   38.196  110.221 211.016
 test1  1.529   2.381   3.527   8.411   40.551  105.16  212.573
 test3  3.035   3.934   8.606   20.858  61.571  118.744 235.428
 test2  3.136   6.238   10.508  33.48   43.532  118.044 239.481
 test10 1.593   4.736   7.527   20.557  59.856  162.907 323.147
 test11 3.913   11.506  23.26   68.644  207.591 600.444 1211.545
```

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
it-hms: Agreed. What is the suggestion?  Can we complete this PR, write a spec... then make another PR.

alexjhawk: This could be completed in another PR, but the suggestion is that we use the StringUtils method, as it has been benchmarked to perform the best. Any other implementation or change of reading a string from an input stream should be separately addressed if we find it necessary.

alexjhawk: I don't think we should introduce code here which performs the same functionality as code we already have and can reference. The referenced method should be addressed (if necessary), instead of creating a new one.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -70,12 +69,29 @@ public static synchronized long getQueueFifoTimeSpanMins() {
...
    * Set the FIFO queue time span in minutes.
    *
+   * @deprecated
    * @param timeSpanMins new FIFO queue time span in minutes
    */
   public static synchronized void setQueueFifoTimeSpanMins(long timeSpanMins) {
     queueFifoTimeSpanMins = timeSpanMins;
   }
 
+  /**
```
alexjhawk, line 83: I thought I had commented on this before, but I don't see it anymore....


I don't think we should be introducing setQueueFifoTimeSpanMinsCheck(). It's redundant and the existing setQueueFifoTimeSpanMins() method should be updated. By changing the method name, it is just adding work in all connectors when they update to this version. 

I actually see this more as a bug fix, because adding the constraint should not affect any existing implementations, and if it does, they were using it improperly to begin with, so the exception may be a good thing so that it can get noticed/caught.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -55,4 +55,10 @@
 
   /** Minimum amount of Historical FIFO queue that can be set by library users. */
   static final long MIN_QUEUE_SPAN_MINS = 1;
+
+  /** The number of milliseconds in one hour. */
+  static final long MILLISECONDS_PER_HOUR = 3600000L;
+
+  /** Default maximum amount of time Historical FIFO can get behind: 48 hours in ms. */
+  static final long MAX_HIST_FIFO_GET_BEHIND_MS = 48L * MILLISECONDS_PER_HOUR;
```
alexjhawk, line 63:
```suggestion
  /** Default maximum amount of time Historical FIFO can get behind: 48 hours in ms. */
  static final long MAX_HIST_FIFO_GET_BEHIND_MS = SCTimeUnit.HOURS.toMillis(48);
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/EBdTimeoutException.java
```diff
@@ -0,0 +1,19 @@
...
+ * @since 1.6.0
+ */
+public class EBdTimeoutException extends Exception {
+
+  /**
+   * Constructor for the EBdTimeoutException.
+   *
+   * @param errorMessage the message to associate with this exception
+   */
+  public EBdTimeoutException(String errorMessage) {
```
alexjhawk, line 16:
```suggestion
public class EbdTimeoutException extends Exception {

  /**
   * Constructor for the EbdTimeoutException.
   *
   * @param errorMessage the message to associate with this exception
   */
  public EbdTimeoutException(String errorMessage) {
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -61,4 +61,10 @@
 
   /** Default maximum amount of time Historical FIFO can get behind: 48 hours in ms. */
   static final long MAX_HIST_FIFO_GET_BEHIND_MS = 48L * MILLISECONDS_PER_HOUR;
+
+  /** Default thread sleep time waiting for EBD in ms. */
+  static final int DEFAULT_EBD_THREAD_SLEEP_MS = 5;
+
+  /** Maximum amount of time that EBD call should wait before timing out. */
+  static final long MAX_EBD_WAIT = 15000;
```
alexjhawk, line 69: Units should be added to tag name -- i.e. MAX_EBD_WAIT_MILLIS

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Ewon Flexy Extensions Library Changelog
```
alexjhawk, line 1: The version number needs to be bumped from 1.5.0 to 1.6.0 in the README.md file.

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: Pepsi challenge shows that execution is about the same. The GC really makes it tricky to profile memory, but StringUtil does appear to use less memory. I'll make the change. 
image

TomKimsey: Is pepsi challenge coming back as a common saying? I've seen it a few places recently.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
TomKimsey: If this is done you can remove the "48 hours in ms." in the comment as the code documents this well.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
it-hms: ok. We can make the change. You are correct, all applications that this effects aren't using it properly.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/EBdTimeoutException.java
it-hms: changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: changed

CHANGELOG.md
it-hms: @alexjhawk  what line ?  it is changed in line 65.  You have this indicated on CHANGELOG.md

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: It never went away for me...

CHANGELOG.md
alexjhawk: Yes, you got it at line 65. I added the comment here to associate with this commit, but couldn't comment on README.md as there we no changes to comment on.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -55,4 +55,10 @@
 
   /** Minimum amount of Historical FIFO queue that can be set by library users. */
   static final long MIN_QUEUE_SPAN_MINS = 1;
+
+  /** The number of milliseconds in one hour. */
+  static final long MILLISECONDS_PER_HOUR = 3600000L;
+
+  /** Default maximum amount of time Historical FIFO can get behind. */
+  static final long MAX_HIST_FIFO_GET_BEHIND_MS = 48L * MILLISECONDS_PER_HOUR;
```
alexjhawk, line 63: See comment at https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/pull/16#discussion_r799516012

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -57,6 +59,175 @@ public static void exportAllHistoricalToFile(
...
+
+  /**
+   * Executes EBD call, waits for data return.
+   *
+   * @param ebdStr EBD string
+   * @return EBD Exporter - caller must close exporter
+   * @throws EbdTimeoutException when there is no response before timeout period
+   * @throws IOException when there is an Exporter Exception
+   */
+  public static Exporter ExecuteEBdCall(String ebdStr) throws IOException, EbdTimeoutException {
```
alexjhawk, line 173: method name is not camel case (should be executeEbdCall)

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/CircularizedFileException.java
```diff
@@ -0,0 +1,12 @@
+package com.hms_networks.americas.sc.extensions.historicaldata;
+/**
```
alexjhawk, line 2: Not sure if it's a hard requirement, but the convention is to include a blank line between package/imports/javadocs.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -443,4 +443,28 @@ private static void updateTrackingStartTime(long updateTime)
...
+   * decision to advance time.
+   *
+   * @throws IOException if unable to read or write files
+   * @throws TimeTrackerUnrecoverableException if both time tracking files are corrupted
+   * @throws CorruptedTimeTrackerException one of the tracking files is corrupted
+   */
+  public static void advanceTrackingStartTime()
+      throws CorruptedTimeTrackerException, IOException, TimeTrackerUnrecoverableException {
+
+    // Get start time from file, or start new time tracker if startNewTimeTracker is true.
```
alexjhawk, line 458: This comment needs to be updated. startNewTimeTracker doesn't exist here

alexjhawk (changes requested):
Mostly small/formatting

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: oh, I see what you mean.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: changed

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/CircularizedFileException.java
it-hms: changed

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
it-hms: changed.

it-hms:
Changed default fallback max to 12 hours. 48 seemed like too much.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -1,7 +1,6 @@
 package com.hms_networks.americas.sc.extensions.historicaldata;
 
 import com.hms_networks.americas.sc.extensions.fileutils.FileAccessManager;
-
 import com.hms_networks.americas.sc.extensions.json.JSONException;
```
TomKimsey, line 4: looks like a whitespace change slipped into this commit

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
```diff
@@ -71,11 +74,31 @@ public static synchronized long getQueueFifoTimeSpanMins() {
...
 
+  /**
+   * Set the maximum amount the Historical FIFO queue can get behind in minutes.
+   *
+   * @param timeMins new FIFO queue time span in minutes
+   * @throws IllegalArgumentException if parameter is not greater than 0
+   */
+  public static synchronized void setQueueMaxBehindMins(long timeMins) {
+    if (timeMins < 1) {
+      throw new IllegalArgumentException("The parameter passed must be greater than 0.");
```
TomKimsey, line 97:
```suggestion
   * @throws IllegalArgumentException if parameter is not greater than or equal to 1
   */
  public static synchronized void setQueueMaxBehindMins(long timeMins) {
    if (timeMins < 1) {
      throw new IllegalArgumentException("The parameter passed must be greater than or equal to 1.");
```
or
```suggestion
   * @throws IllegalArgumentException if parameter is not greater than 0
   */
  public static synchronized void setQueueMaxBehindMins(long timeMins) {
    if (timeMins <= 0) {
      throw new IllegalArgumentException("The parameter passed must be greater than 0.");
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -52,4 +54,16 @@
...
+  static final long MIN_QUEUE_SPAN_MINS = 1;
+
+  /** Default maximum amount of time Historical FIFO can get behind. */
+  static final long MAX_HIST_FIFO_GET_BEHIND_MS = SCTimeUnit.HOURS.toMillis(12);
+
+  /** Default thread sleep time waiting for EBD in ms. */
+  static final int DEFAULT_EBD_THREAD_SLEEP_MS = 5;
+
+  /** Maximum amount of time that EBD call should wait in milliseconds before timing out. */
+  static final long MAX_EBD_WAIT_MILLIS = 15000;
```
TomKimsey, line 68: To be consist ant with the other constants. 

```suggestion
  static final long MAX_EBD_WAIT_MS = 15000;
```

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -57,6 +59,175 @@ public static void exportAllHistoricalToFile(
...
+
+  /**
+   * Executes EBD call, waits for data return.
+   *
+   * @param ebdStr EBD string
+   * @return EBD Exporter - caller must close exporter
+   * @throws EbdTimeoutException when there is no response before timeout period
+   * @throws IOException when there is an Exporter Exception
+   */
+  public static Exporter executeEBdCall(String ebdStr) throws IOException, EbdTimeoutException {
```
alexjhawk, line 173: Method name still capitalized incorrectly. B should be lowercase.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
```diff
@@ -346,6 +346,8 @@ public static ArrayList parseHistoricalFile(String filename) throws IOException,
       // Read next line before looping again
       line = reader.readLine();
     }
+    reader.close();
```
alexjhawk, line 358: Should we even keep this old EBD to file code? I think it should probably just be removed to reduce confusion.

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
it-hms: yeah, that was from formatting... will move into the formatting commit

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
it-hms: ok... this is like Sergio's 1+1 != 2

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: Agreed. but I never did have the chance to test strings.... I got nervous when I had to reformat the Flexy.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
TomKimsey: Yeah, there is nothing technically wrong with how you had it. Its just better code style to match the value in the log and comments to the value being used in the functional portion of the code.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java
it-hms: changed.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueManager.java
TomKimsey: You could even have this which would make it so that any change to the value would require just one line change. 
```suggestion
   * @throws IllegalArgumentException if timeMin parameter is invalid
   */
  public static synchronized void setQueueMaxBehindMins(long timeMins) {
    int minimumTimeMin = 0;
    if (timeMins <= minimumTimeMin) {
      throw new IllegalArgumentException("The parameter passed must be greater than " + minimumTimeMin + ".");
```

alexjhawk: > You could even have this which would make it so that any change to the value would require just one line change.

While this simplifies the code slightly and is less work if the value needs to change, I think it is much less user friendly. Users' IDEs may suggest Javadocs to them as they develop, and by removing the number from the Javadocs, the developer either has to guess on the constraint and hope they don't get an exception or dig through our source code. For example, this is what I can see in IntelliJ for the parseHistoricalFile() method.

image

TomKimsey: Probably would be better to just make it final and use @link then.

alexjhawk (approved):
Did not see anything else. I think updating the string EBD logic to match the regular EBD logic should be a priority though because many connectors use both regular and string data and I don't think they should be using two different EBD mechanisms.

TomKimsey (approved):

PR #17, opened 2022-02-21
Dev/1.7.0

alexjhawk:


alexjhawk:
@TomKimsey @oliver-walker-hms @it-hms This is ready for review and has been verified/tested against Cumulocity connector.

src/main/java/com/hms_networks/americas/sc/extensions/util/Base64.java
```diff
@@ -0,0 +1,2597 @@
+package com.hms_networks.americas.sc.extensions.util;
+
```
it-hms, line 2: Please add a comment that this class is copyright Rob Harder and has been deemed to belong in the public domain. 
```
The source code for this class was graciously provided to the public domain by Robert Harder. He shall retain all copyrights. 
```

src/main/java/com/hms_networks/americas/sc/extensions/util/Base64.java
```diff
@@ -0,0 +1,2597 @@
...
+ * </ul>
+ *
+ * <p>I am placing this code in the Public Domain. Do with it as you will. This software comes with
+ * no guarantees or warranties but with plenty of well-wishing instead! Please visit <a
+ * href="http://iharder.net/base64">http://iharder.net/base64</a> periodically to check for updates
+ * or to contribute improvements.
+ *
+ * @author Robert Harder
+ * @author rob@iharder.net
+ * @author HMS Networks, MU Americas Solution Center
```
it-hms, line 131: Did we add anything?

it-hms (changes requested):
I think  we should add a comment about the copyright of the code. Also, we should remove the @author javadoc if we did not author anything.  Not sure if this was added by jetbrains or how this was added.

src/main/java/com/hms_networks/americas/sc/extensions/util/Base64.java
alexjhawk: Rob Harder has already added a comment pertinent to this including his website and contact information: https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/blob/dev/1.7.0/src/main/java/com/hms_networks/americas/sc/extensions/util/Base64.java#L124-L130

src/main/java/com/hms_networks/americas/sc/extensions/util/Base64.java
alexjhawk: We made it compatible with Java 1.4 -- See https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/blob/dev/1.7.0/src/main/java/com/hms_networks/americas/sc/extensions/util/Base64.java#L31-L33

alexjhawk: Changes mainly consisted of removing String.format, @ Override annotations and type parameters.

TomKimsey: I think this is a arguable case for adding a non-buildable commit. For tracking sake it would be nice if there was a commit to add the file directly as you found it, then the next commit being the porting for ewon java 1.4. That way all the changes we made are easily trackable in the event something doesn't work right or we port to new hardware or the original source is updated.

alexjhawk: @TomKimsey Added separate commit(s)

it-hms (approved):

oliver-walker-hms (approved):

src/main/java/com/hms_networks/americas/sc/extensions/util/Base64.java
```diff
@@ -28,6 +28,9 @@
  * <p>Change Log:
  *
  * <ul>
+ *   <li>v2.3.7_J4 - Add support for Java 4 by removing all instances of <code>String.format()
+ *       </code>, <code>@Override</code>, and type parameters. (Author: HMS Networks, MU Americas
+ *       Solution Center)
```
TomKimsey, line 33: Should this update commit just remove this changelog list? We have not released any of the previous versions of this and none of our other classes have changelogs.

alexjhawk: We could remove it. I'm not really sure of any sort of conventions for this scenario. In general in this PR though, I tried to modify as little as possible to respect the developer who put it into the public domain, but also clearly indicate that we did modify it, and what we modified.

alexjhawk: I might look to @it-hms for a definitive answer on a decision like this, but I think leaving it and contributing our own changelog entry here allows us to clearly identify the changes we have made and retain the established format.

it-hms: I like it. It give us copyright protections.

TomKimsey (approved):

src/main/java/com/hms_networks/americas/sc/extensions/util/Base64.java
TomKimsey: Sounds good!

PR #21, opened 2022-09-30
Dev/1.10.0

alexjhawk:


oliver-walker-hms (approved):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppArgsParser.java
```diff
@@ -0,0 +1,62 @@
+package com.hms_networks.americas.sc.extensions.system.application;
+
+/**
+ * A utility class for parsing the arguments passed to the main method of a Solution Center Java
```
TomKimsey, line 4: Why is the language making it seem that this can only be used for solution center applications? Isnt this really open to anyone to use.

alexjhawk: There is no way to pass arguments to an Ewon Java application at this point unless you are using the solution center multi-loader application. I figured labeling would help clarify, as this class is not useful for applications outside of those which support our multi-loader application.

TomKimsey: image

Thats not really what the readme leads me to believe the name is. Sure the repo  is "sc-flexy-java-multi-loader" but thats really the only reference i see to "SC" in the repo. I dont see solution center really mentioned in full anywhere.

alexjhawk: Will make changes to genericize comments.

alexjhawk: Updated

TomKimsey (approved):

PR #22, opened 2022-10-20
Add application pause while WAN initializes

oliver-walker-hms:


src/main/java/com/hms_networks/americas/sc/extensions/system/time/SCTimeUnit.java
```diff
@@ -11,19 +11,19 @@
  */
```
alexjhawk, line 11: Improper usage of this API class. This commit should be dropped and constants remain private

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+    return;
+  }
+
+  /**
+   * Helper function to pause application for set period of time while a WAN connection is being
+   * established.
+   */
+  private static void waitForWanHelper() {
+    int seconds = 20;
+    long sleepTimeMillis = seconds * SCTimeUnit.MILLISECONDS_PER_SECOND;
```
alexjhawk, line 145:
```suggestion
    long sleepTimeMillis = SCTimeUnit.SECONDS.toMillis(seconds);
```

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+   * @return true if an initialized WAN IP address is found.
+   */
+  public static boolean deviceHasWanIP() {
+    boolean hasWanIp = false;
+    final String uninitializedWan = "0.0.0.0";
+
+    if (NetManager.getWanIp().equalsIgnoreCase(uninitializedWan)) {
+      hasWanIp = true;
+    }
+    return hasWanIp;
```
alexjhawk, line 74:
```suggestion
    final String uninitializedWan = "0.0.0.0";
    return NetManager.getWanIp().equalsIgnoreCase(uninitializedWan);
```

alexjhawk: This method can be simplified

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+  /**
+   * Waits for a number of minutes for a WAN IP to be establlshed if one does not already exist.
+   * After the given time has passed, returns false indicating no WAN IP is present.
+   *
+   * @param timeOutMinutes returns true if an initialized WAN IP is detected
+   */
+  public static boolean waitForWanIp(int timeOutMinutes) {
+
+    // There are 20 seconds in a wait period. 3 periods exist in a minute
+    int minutesToPeriodsConverter = 3;
```
alexjhawk, line 86: I think the variable `minutesToPeriodsConverter` may be more appropriately named something like `wanIpChecksPerMinute` or even just `periodsPerMinute`.

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+              "No WAN connection detected, please ensure there is a WAN connection before starting"
+                  + " the connector.");
+          wanConnectionEstablished = false;
+          i = maxWaitPeriods;
+        }
+        // Wait 20 seconds on WAN connection initialization
+        waitForWanHelper();
+
+      } else {
+        Logger.LOG_INFO("WAN connection found.");
```
alexjhawk, line 107: Logging should not be included in the library itself. The `Wan connection found` log is likely redundant as WAN connection should be expected before continuing, and if required could be shown by the connector itself when the return value was `true`.

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+    // Ensure there is a valid WAN IP address
+    for (int i = 0; i < maxWaitPeriods; i++) {
+      Logger.LOG_INFO("WAN IP is: " + NetManager.getWanIp());
+
+      if (deviceHasWanIP()) {
+        // If no connection is established by end of wait time
+        if (i == (maxWaitPeriods - 1)) {
+          Logger.LOG_CRITICAL(
+              "No WAN connection detected, please ensure there is a WAN connection before starting"
+                  + " the connector.");
```
alexjhawk, line 99: Logging should not be included in the library itself.

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+   */
+  public static void waitForWanIp() {
+    boolean continueLooping = true;
+    // Ensure there is a valid WAN IP address
+    while (continueLooping) {
+      Logger.LOG_INFO("WAN IP is: " + NetManager.getWanIp());
+
+      if (!deviceHasWanIP()) {
+        // Wait 20 seconds on WAN connection initialization
+        waitForWanHelper();
```
alexjhawk, line 128:
```suggestion
        // Wait 20 seconds on WAN connection initialization
        final int waitTimeSeconds = 20;
        waitForWanHelper(waitTimeSeconds);
```

alexjhawk: The inline comment here refers to `Wait 20 seconds` but the 20-second constant is not defined in this method. 

I think that the seconds value should just get passed down into waitForWanHelper(...) as it makes the solution more flexible as there is only one point to change relevant information (the variable and the comment)

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+
+    return;
+  }
+
+  /**
+   * Helper function to pause application for set period of time while a WAN connection is being
+   * established.
+   */
+  private static void waitForWanHelper() {
+    int seconds = 20;
```
alexjhawk, line 144:
```suggestion
  private static void waitForWanHelper(int seconds) {
```

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+        // Wait 20 seconds on WAN connection initialization
+        waitForWanHelper();
+
+      } else {
+        Logger.LOG_INFO("WAN connection found.");
+        continueLooping = false;
+      }
+    }
+
+    return;
```
alexjhawk, line 136:
```suggestion
```
Unnecessary `return;` in `void` function

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+    }
+
+    return wanConnectionEstablished;
+  }
+
+  /**
+   * Pause current thread indefinitely if an initialized WAN IP has not been established. When a WAN
+   * IP is detected, resume application.
+   */
+  public static void waitForWanIp() {
```
alexjhawk, line 120: Same comments as above about using the logging functionality within the library itself. 

In the library, in most cases, a successful result should simply be indicated by returning a boolean `true` or just getting no error. If there is an error, the library should throw an exception.

This architecture allows any application which uses the library to decide how it handles errors in the manner it sees fit, which provides flexibility overall.

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+        // If no connection is established by end of wait time
+        if (i == (maxWaitPeriods - 1)) {
+          Logger.LOG_CRITICAL(
+              "No WAN connection detected, please ensure there is a WAN connection before starting"
+                  + " the connector.");
+          wanConnectionEstablished = false;
+          i = maxWaitPeriods;
+        }
+        // Wait 20 seconds on WAN connection initialization
+        waitForWanHelper();
```
alexjhawk, line 104:
```suggestion
        // Wait 20 seconds on WAN connection initialization
        final int waitTimeSeconds = 20;
        waitForWanHelper(waitTimeSeconds);
````

Same as below suggestion

File: src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java [hunk 12]
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
+
+  /**
+   * Waits for a number of minutes for a WAN IP to be establlshed if one does not already exist.
+   * After the given time has passed, returns false indicating no WAN IP is present.
+   *
+   * @param timeOutMinutes returns true if an initialized WAN IP is detected
+   */
+  public static boolean waitForWanIp(int timeOutMinutes) {
+
+    // There are 20 seconds in a wait period. 3 periods exist in a minute
```
alexjhawk, line 85: This comment refers to the 20 seconds in a period, which isn't yet defined.  See comment at line 128

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +58,98 @@ public static void disableAppAutoRestart() {
...
   }
+
+  /**
+   * Checks if a valid WAN IP address has been assigned to the device.
+   *
+   * @return true if an initialized WAN IP address is found.
+   */
+  public static boolean deviceHasWanIP() {
+    boolean hasWanIp = false;
+    final String uninitializedWan = "0.0.0.0";
```
alexjhawk, line 69:
```suggestion
    final String uninitializedWanIp = "0.0.0.0";
```
This variable could be named a bit more consistently as it is compared with .getWanIp()

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Ewon Flexy Extensions Library Changelog
 
+## Version 1.11.0
+### Features
+- Added support for application pause on no valid WAN
```
alexjhawk, line 5:
```suggestion
- Added support for application pause when no WAN IP address is available
```

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
alexjhawk: After writing this comment, I realized that the methods directly above your additions also contain Logger.* calls. It's not relevant to this PR, but I would also like to see those Logger calls removed in the future. Architecturally, I really think it is up to the application itself to be handling logging.

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +61,72 @@ public static void disableAppAutoRestart() {
...
+    return NetManager.getWanIp().equalsIgnoreCase(uninitializedWanIp);
+  }
+
+  /**
+   * Waits for a number of minutes for a WAN IP to be establlshed if one does not already exist.
+   * After the given time has passed, returns false indicating no WAN IP is present.
+   *
+   * @param timeOutMinutes returns true if an initialized WAN IP is detected
+   */
+  public static boolean waitForWanIp(int timeOutMinutes) throws InterruptedException {
```
alexjhawk, line 81: What are your thoughts on making this seconds instead of minutes? 

I feel like down the road we might end up wanting/needing to tweak this value. If we establish that waitForWanIp(*int*) is minutes-based, it will be challenging to switch to seconds later to allow for better precision, as that would break any existing usages/calls to this method.

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +61,72 @@ public static void disableAppAutoRestart() {
...
+  public static boolean deviceHasWanIP() {
+    final String uninitializedWanIp = "0.0.0.0";
+    return NetManager.getWanIp().equalsIgnoreCase(uninitializedWanIp);
+  }
+
+  /**
+   * Waits for a number of minutes for a WAN IP to be establlshed if one does not already exist.
+   * After the given time has passed, returns false indicating no WAN IP is present.
+   *
+   * @param timeOutMinutes returns true if an initialized WAN IP is detected
```
alexjhawk, line 79: incorrect javadoc

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
alexjhawk: For example, this would make the method super flexible and allow any time unit: 

```
  /**
   * Waits for a number of minutes for a WAN IP to be established if one does not already exist.
   * After the given time has passed, returns false indicating no WAN IP is present.
   *
   * @param timeout the timeout value in the specified time unit
   * @param timeUnit the time unit of the timeout value
   * @return true if an initialized WAN IP is detected, false otherwise
   */
  public static boolean waitForWanIp(int timeout, SCTimeUnit timeUnit) throws InterruptedException {
    // Convert timeout to seconds
    long timeoutSeconds = timeUnit.toSeconds(timeout);

    // Get rounded up number of wait periods in timeout (double cast to avoid integer division)
    int maxWaitPeriods = (int) Math.ceil((double) timeoutSeconds / (double) secondsInWaitPeriod);
```

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
alexjhawk: I suggested correct javadocs in the above comment https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/pull/22#discussion_r1002152290

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
oliver-walker-hms: I like this solution a lot better. The time unit flexibility is much more ideal as an API method

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -56,4 +61,72 @@ public static void disableAppAutoRestart() {
...
+        // If no connection is established by end of wait time
+        if (i == (maxWaitPeriods - 1)) {
+          wanConnectionEstablished = false;
+          i = maxWaitPeriods;
+        } else {
+          waitForWanHelper(SECONDS_IN_WAIT_PERIOD);
+        }
+      } else {
+        wanConnectionEstablished = true;
+        i = maxWaitPeriods;
```
alexjhawk, line 104:
```suggestion
        break;
```
Alternatively, you could just use `break;` here to break out of the loop.

alexjhawk (approved):

oliver-walker-hms:
I'm going to test this one more time Monday morning before merging

PR #27, opened 2022-11-11
Dev/1.12.1

alexjhawk:


TomKimsey (commented):
Were the crashes happening because the preallocated space was too small? There are performance reasons for why this exists and does not rely on dynamic memory allocation.

TomKimsey:
> > Were the crashes happening because the preallocated space was too small? There are performance reasons for why this exists and does not rely on dynamic memory allocation.
> 
> Yes, and I've previously increased the limit, but it is happening again in a different location. As for performance benefits, what is there besides the pre-allocation of memory? A native Java StringBuffer will pre-allocate the memory for a string of the specified length. The only difference I'm aware of is that if the limit is reached, it will expand instead of an error.

Ill have to look back into my notes on what the benefit was. This was years ago at this point. I recall stringbuffer having undesirable behavior.

oliver-walker-hms:
I remember there may have been issues when string became very large and that was why we had to have pre-allocated string builder. I don't remember what those issues were.

TomKimsey:
> I remember there may have been issues when strings became very large and that was why we had to have pre-allocated string builder. I don't remember what those issues were.

It might just be that as long as you choose a capacity that should not be exceeded you are fine. 
StringBuffer is thread safe and as such has the performance hit of the synchronization tasks it has to perform.

alexjhawk:
It would be great to see the notes/reasons from before because I couldn't remember. In terms of testing, I didn't encounter any issues during some initial OSIsoft connector testing.

I agree that the best scenario is one where all of the desired memory is pre-allocated and no upper limit is reached, but I think this is sort of wishful thinking and sets us up for failure/bugs.

TomKimsey:
> It would be great to see the notes/reasons from before because I couldn't remember. In terms of testing, I didn't encounter any issues during some initial OSIsoft connector testing.
> 
> I agree that the best scenario is one where all of the desired memory is pre-allocated and no upper limit is reached, but I think this is sort of wishful thinking and sets us up for failure/bugs.

Straight from the docs:

> The StringBuilder class should generally be used in preference to this one, as it supports all of the same operations but it is faster, as it performs no synchronization.

https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuffer.html

alexjhawk:
> > I remember there may have been issues when strings became very large and that was why we had to have pre-allocated string builder. I don't remember what those issues were.
> 
> It might just be that as long as you choose a capacity that should not be exceeded you are fine. StringBuffer is thread safe and as such has the performance hit of the synchronization tasks it has to perform.

This is my understanding. Ideally, we would be choosing a capacity that will work for most cases, but also not setting a hard limit which can bring a whole connector down, because, for example, a customer has an extremely long tag name.

alexjhawk:
> > It would be great to see the notes/reasons from before because I couldn't remember. In terms of testing, I didn't encounter any issues during some initial OSIsoft connector testing.
> > I agree that the best scenario is one where all of the desired memory is pre-allocated and no upper limit is reached, but I think this is sort of wishful thinking and sets us up for failure/bugs.
> 
> Straight from the docs:
> 
> > The StringBuilder class should generally be used in preference to this one, as it supports all of the same operations but it is faster, as it performs no synchronization.
> 
> https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuffer.html

Sadly, StringBuilder is not available in Java 1.4, but I don't think synchronization is probably too big of an impact. I also don't think having synchronization is a bad thing, because is some connectors, we could be calling the payload building methods from different data retrieval threads.

TomKimsey:
> > > It would be great to see the notes/reasons from before because I couldn't remember. In terms of testing, I didn't encounter any issues during some initial OSIsoft connector testing.
> > > I agree that the best scenario is one where all of the desired memory is pre-allocated and no upper limit is reached, but I think this is sort of wishful thinking and sets us up for failure/bugs.
> > 
> > 
> > Straight from the docs:
> > > The StringBuilder class should generally be used in preference to this one, as it supports all of the same operations but it is faster, as it performs no synchronization.
> > 
> > 
> > https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuffer.html
> 
> Sadly, StringBuilder is not available in Java 1.4, but I don't think synchronization is probably too big of an impact. I also don't think having synchronization is a bad thing, because is some connectors, we could be calling the payload building methods from different data retrieval threads.

Thats why we wrote our own :)

alexjhawk:
> > > > It would be great to see the notes/reasons from before because I couldn't remember. In terms of testing, I didn't encounter any issues during some initial OSIsoft connector testing.
> > > > I agree that the best scenario is one where all of the desired memory is pre-allocated and no upper limit is reached, but I think this is sort of wishful thinking and sets us up for failure/bugs.
> > > 
> > > 
> > > Straight from the docs:
> > > > The StringBuilder class should generally be used in preference to this one, as it supports all of the same operations but it is faster, as it performs no synchronization.
> > > 
> > > 
> > > https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuffer.html
> > 
> > 
> > Sadly, StringBuilder is not available in Java 1.4, but I don't think synchronization is probably too big of an impact. I also don't think having synchronization is a bad thing, because is some connectors, we could be calling the payload building methods from different data retrieval threads.
> 
> Thats why we wrote our own :) "PreAllocatedStringBuilder.java" used to be called "StringBuilderLite.java"



> > > > I remember there may have been issues when strings became very large and that was why we had to have pre-allocated string builder. I don't remember what those issues were.
> > > 
> > > 
> > > It might just be that as long as you choose a capacity that should not be exceeded you are fine. StringBuffer is thread safe and as such has the performance hit of the synchronization tasks it has to perform.
> > 
> > 
> > This is my understanding. Ideally, we would be choosing a capacity that will work for most cases, but also not setting a hard limit which can bring a whole connector down, because, for example, a customer has an extremely long tag name.
> 
> log the exception and tag name lenth, add a config parameter "MaxTagNameLength" that can override the default. Or you could even have the connector modify the config file itself, and restart with the updated size.
> 
> Have you tested this change on a connector with 500+ tags? You may see the performance hit of using StringBuffer there.

I have not tested this with 500+ tags. 

At this point, maybe the class remains mostly as is, but I would greatly largely prefer not to introduce even more configuration file parameters. In my opinion, we tend to push most things to the configuration file, but there is no reason for us to burden a user in this case.

I would rather see an update to the `PreAllocatedStringBuilder` class to automatically expand itself as StringBuffer would than retain hard limits in code. The fewer the hard limits, the better because it's one less point of failure from both a customer and developer point of view. I'm currently dealing with the second customer-reported bug because of a hard limit imposed via this class. I don't think it promotes a good user experience to continue with this implementation as it currently is.

TomKimsey:
> I have not tested this with 500+ tags.
> 
> At this point, maybe the class remains mostly as is, but I would greatly largely prefer not to introduce even more configuration file parameters. In my opinion, we tend to push most things to the configuration file, but there is no reason for us to burden a user in this case.
> 
> I would rather see an update to the `PreAllocatedStringBuilder` class to automatically expand itself as StringBuffer would than retain hard limits in code. The fewer the hard limits, the better because it's one less point of failure from both a customer and developer point of view. I'm currently dealing with the second customer-reported bug because of a hard limit imposed via this class. I don't think it promotes a good user experience to continue with this implementation as it currently is.

It would be nice to see the performance implications of this in a system with a high tag count. I suspect there would be issues and without that information I dont want to change this class right away.

 Could you point to the code thats failing for the customer? I would like to take a look and see how its being handled.

alexjhawk:
> > I have not tested this with 500+ tags.
> > At this point, maybe the class remains mostly as is, but I would greatly largely prefer not to introduce even more configuration file parameters. In my opinion, we tend to push most things to the configuration file, but there is no reason for us to burden a user in this case.
> > I would rather see an update to the `PreAllocatedStringBuilder` class to automatically expand itself as StringBuffer would than retain hard limits in code. The fewer the hard limits, the better because it's one less point of failure from both a customer and developer point of view. I'm currently dealing with the second customer-reported bug because of a hard limit imposed via this class. I don't think it promotes a good user experience to continue with this implementation as it currently is.
> 
> It would be nice to see the performance implications of this in a system with a high tag count. I suspect there would be issues and without that information I dont want to change this class right away.
> 
> Could you point to the code thats failing for the customer? I would like to take a look and see how its being handled.

I don't see why we couldn't just update our implementation to support automatically growing in size. It has the potential to improve the user experience, and considering that the alternative is to simply throw an error at the hard limit, it shouldn't make a big difference. 
The worst case scenario is that the system runs low on memory (which we should catch and warn about when we already check for new data).
The best case scenario is that the connector continues working and only has a temporary impact to performance while growing the underlying array size (and of course the corresponding memory increase).

The first bug was resolved via this commit (still enforcing a hard limit): https://github.com/hms-networks/flexy-osisoft-connector/commit/27035987f85f626a085f80cdd93b54204c4c2d36

The second bug is happening at the following line: https://github.com/hms-networks/flexy-osisoft-connector/blob/30b7fca5aeae71e899ab444a4f4c699770f732dd/src/main/java/com/hms/flexyosisoftconnector/payloadhandler/OsisoftJsonPayload.java#L384

TomKimsey:
> > > I have not tested this with 500+ tags.
> > > At this point, maybe the class remains mostly as is, but I would greatly largely prefer not to introduce even more configuration file parameters. In my opinion, we tend to push most things to the configuration file, but there is no reason for us to burden a user in this case.
> > > I would rather see an update to the `PreAllocatedStringBuilder` class to automatically expand itself as StringBuffer would than retain hard limits in code. The fewer the hard limits, the better because it's one less point of failure from both a customer and developer point of view. I'm currently dealing with the second customer-reported bug because of a hard limit imposed via this class. I don't think it promotes a good user experience to continue with this implementation as it currently is.
> > 
> > 
> > It would be nice to see the performance implications of this in a system with a high tag count. I suspect there would be issues and without that information I dont want to change this class right away.
> > Could you point to the code thats failing for the customer? I would like to take a look and see how its being handled.
> 
> I don't see why we couldn't just update our implementation to support automatically growing in size. It has the potential to improve the user experience, and considering that the alternative is to simply throw an error at the hard limit, it shouldn't make a big difference. The worst case scenario is that the system runs low on memory (which we should catch and warn about when we already check for new data). The best case scenario is that the connector continues working and only has a temporary impact to performance while growing the underlying array size (and of course the corresponding memory increase).
> 
> The first bug was resolved via this commit (still enforcing a hard limit): [hms-networks/flexy-osisoft-connector@2703598](https://github.com/hms-networks/flexy-osisoft-connector/commit/27035987f85f626a085f80cdd93b54204c4c2d36)
> 
> The second bug is happening at the following line: https://github.com/hms-networks/flexy-osisoft-connector/blob/30b7fca5aeae71e899ab444a4f4c699770f732dd/src/main/java/com/hms/flexyosisoftconnector/payloadhandler/OsisoftJsonPayload.java#L384

Ill take a look at the usage and changes you linked. My initial thought is that the implementation should make sure to allocate only the space needed to extend the array (by adding additional arrays) rather than allocating just a larger array then transferring the contents.

Im a bit confused why in this codebase the return value of .append is never checked.

alexjhawk:
Another additional benefit to this change is that when applications such as connectors are updated to directly use `StrBuilder`, a significant number of new methods will be available, including constructors.

Examples:
1. Append more than String data type (`append(String)`, `append(int)`, `append(float)`, etc)
2. Append using chained-expression calls (`append("Word of the day is ").append(wordOfTheDay).append("!");`)
3. Append as line, adds new line after automatically (`appendln("New line after this");`)
4. Get a `Reader` of the builder's contents (`asReader()`) or a `StrTokenizer` to tokenize the contents of the builder (`asTokenizer()`)
5. etc...

TomKimsey:
@alexjhawk have you characterized the performance vs the current library? It sounds like there shouldn't be much if any change but I would hate to introduce unintentional performance losses.

it-hms (approved):

alexjhawk:
> @alexjhawk have you characterized the performance vs the current library? It sounds like there shouldn't be much if any change but I would hate to introduce unintentional performance losses.

I have not done any performance testing for this, although I did verify functionality using two different connectors. In terms of performance, I would expect that it would be identical or slightly better because in terms of implementation, the only differences mostly are with variable names, and of course the addition of the `ensureCapacity()` check, which is a single if statement that is bypassed unless the underlying array is too small.

I can make a short test right now to verify though.

alexjhawk:
> > @alexjhawk have you characterized the performance vs the current library? It sounds like there shouldn't be much if any change but I would hate to introduce unintentional performance losses.
> 
> I have not done any performance testing for this, although I did verify functionality using two different connectors. In terms of performance, I would expect that it would be identical or slightly better because in terms of implementation, the only differences mostly are with variable names, and of course the addition of the `ensureCapacity()` check, which is a single if statement that is bypassed unless the underlying array is too small.
> 
> I can make a short test right now to verify though.

@TomKimsey I have created a short test to characterize the performance of a few scenarios:
1. `StrBuilder` without specifying a size (unknown size)
2. `StrBuilder` with a specified size (known size)
3. `PreAllocatedStringBuilder` with a specified size (known size)

The results were nearly identical across the board for a `StrBuilder` and `PreAllocatedStringBuilder` with known/specified size. As expected, a `StrBuilder` without specifying the size takes a performance hit due to the underlying array resize (but it doesn't error out like `PreAllocatedStringBuilder`).

I also don't have an exact explanation for why the first test took a performance hit in most tests, but I assume it has to do with memory allocation during the first run.

Verbose Results:
```
TEST EXECUTION 2022-11-17T13:57:29.550-05:00

Random string 0 is sSa70wUAwd
Random string 1 is 76ouhKUX9v
Random string 2 is c3i8isNDaB
Random string 3 is SmJ81AYiBB
Random string 4 is 7MTQqaQo6E
Random string 5 is FERH3vJsNX
Random string 6 is PmQ03dHMik
Random string 7 is J5v8OIU1MF
Random string 8 is WTgC5nu0Ef
Random string 9 is B3rsFcFBgo
Random string 10 is ELu8qFYHQf
Random string 11 is dlpvwEexvR
Random string 12 is oSRvfhnxCd
Random string 13 is uUJ0OO80jY
Random string 14 is c3SV2AyZOk
Random string 15 is huY3lccnKi
Random string 16 is qwHkqUA9gz
Random string 17 is Sw5R9vloLs
Random string 18 is ht6HKbGHjo
Random string 19 is pOdLoPFlF6
Random string 20 is sJBv0HkJ4M
Random string 21 is n75p4azidf
Random string 22 is UR6xo1lMOu
Random string 23 is MR7GfRIEo5
Random string 24 is TbuM5vcuAd
Random string 25 is RcmGrGd0oe
Random string 26 is 7aPg0i5uZH
Random string 27 is bh4dPL8m4K
Random string 28 is tTsyDuuvq4
Random string 29 is KsDryBTQul
Random string 30 is KMAXeB6cBk
Random string 31 is EonMMKdBbx
Random string 32 is PuDHpYQFZf
Random string 33 is N7vqUwNy31
Random string 34 is dCxZZGy6lO
Random string 35 is IpoD3QoxFd
Random string 36 is h81uK1yr3Q
Random string 37 is PM7U9gKPYM
Random string 38 is wxLRRLTaTX
Random string 39 is V4WyelineR
Random string 40 is R1HRc5UkyD
Random string 41 is F5SFSoQA6E
Random string 42 is nQrrcNRi7Q
Random string 43 is WHPtGtGlp0
Random string 44 is zjt7Tu0wlZ
Random string 45 is MGafZBnQkr
Random string 46 is 7PKR7UFZtU
Random string 47 is EMFsFIBHe0
Random string 48 is zOhr8h8VXy
Random string 49 is Ubwgm9bRfS
Random string 50 is 7PQ5Y5QjMO
Generated 51 random strings of length 10 for a total of 510 bytes.

Str builder concatenation took 87ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 3ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 14ms with a known size of 510. Expected size: 510
Str builder concatenation took 2ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
Str builder concatenation took 2ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
Str builder concatenation took 3ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
Str builder concatenation took 3ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
Str builder concatenation took 2ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
Str builder concatenation took 3ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
Str builder concatenation took 2ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
Str builder concatenation took 2ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
Str builder concatenation took 3ms with an unknown size of 510. Expected size: 510
Str builder concatenation took 2ms with a known size of 510. Expected size: 510
PreAllocatedStrBuilder concatenation took 2ms with a known size of 510. Expected size: 510
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Ewon Flexy Extensions Library Changelog
 
+## Version 1.12.1
+### Features
+- N/A
+### Bug Fixes
+- N/A
+### Other
+- Deprecated the PreAllocatedStringBuilder.java class in favor of the Java native StringBuffer class
```
TomKimsey, line 9:
```suggestion
- Deprecated the PreAllocatedStringBuilder.java class in favor of the Apache Commons Lang library StrBuilder class (org.apache.commons.lang.text.StrBuilder)
```

alexjhawk:
If curious, the actual test code is:
<details>
```
    setLocalTimeZone();

    // Test performance of strings
    String reportFile = "/usr/StringConcatPerformanceReport.txt";
    String log;
    log =
        "\r\nTEST EXECUTION "
            + SCTimeUtils.getIso8601LocalTimeFormat().format(new Date())
            + "\r\n\r\n";
    Logger.LOG_SERIOUS(log);
    FileAccessManager.appendStringToFile(reportFile, log);
    // Generate an array of size 50-100 of random strings
    final int randomStringLength = 10;
    final int randomStringCount = 50 + new Random().nextInt(50);
    final int totalExpectedBytes = randomStringCount * randomStringLength;
    String[] randomStrings = new String[randomStringCount];
    for (int i = 0; i < randomStringCount; i++) {
      // Add random string with random characters of random length
      randomStrings[i] = RandomStringUtils.random(randomStringLength, true, true);
      log = "Random string " + i + " is " + randomStrings[i] + "\r\n";
      Logger.LOG_SERIOUS(log);
      FileAccessManager.appendStringToFile(reportFile, log);
    }
    log =
        "Generated "
            + randomStringCount
            + " random strings of length "
            + randomStringLength
            + " for a total of "
            + totalExpectedBytes
            + " bytes."
            + "\r\n\r\n";
    Logger.LOG_SERIOUS(log);
    FileAccessManager.appendStringToFile(reportFile, log);
    for (int testNum = 0; testNum < 10; testNum++) {
      log = "Test Number #" + testNum + "\r\n\r\n";
      Logger.LOG_SERIOUS(log);

      // Test performance of str builder (unknown size)
      long startTime = System.currentTimeMillis();
      StrBuilder strBuilder = new StrBuilder();
      for (int i = 0; i < randomStrings.length; i++) {
        strBuilder.append(randomStrings[i]);
      }
      long endTime = System.currentTimeMillis();
      log =
          "Str builder concatenation took "
              + (endTime - startTime)
              + "ms with an unknown size of "
              + strBuilder.length()
              + ". Expected size: "
              + totalExpectedBytes
              + "\r\n";
      Logger.LOG_SERIOUS(log);
      FileAccessManager.appendStringToFile(reportFile, log);

      // Test performance of str builder (known size)
      startTime = System.currentTimeMillis();
      strBuilder = new StrBuilder(totalExpectedBytes);
      for (int i = 0; i < randomStrings.length; i++) {
        strBuilder.append(randomStrings[i]);
      }
      endTime = System.currentTimeMillis();
      log =
          "Str builder concatenation took "
              + (endTime - startTime)
              + "ms with a known size of "
              + strBuilder.length()
              + ". Expected size: "
              + totalExpectedBytes
              + "\r\n";
      Logger.LOG_SERIOUS(log);
      FileAccessManager.appendStringToFile(reportFile, log);

      // Test performance of PreAllocatedStrBuilder (known size)
      startTime = System.currentTimeMillis();
      PreAllocatedStringBuilder preAllocatedStrBuilder =
          new PreAllocatedStringBuilder(totalExpectedBytes);
      for (int i = 0; i < randomStrings.length; i++) {
        preAllocatedStrBuilder.append(randomStrings[i]);
      }
      endTime = System.currentTimeMillis();
      log =
          "PreAllocatedStrBuilder concatenation took "
              + (endTime - startTime)
              + "ms with a known size of "
              + preAllocatedStrBuilder.getLength()
              + ". Expected size: "
              + totalExpectedBytes
              + "\r\n";
      Logger.LOG_SERIOUS(log);
      FileAccessManager.appendStringToFile(reportFile, log);
    }
    System.exit(0);
```
</details>

CHANGELOG.md
alexjhawk: fixed - Can't believe I missed that 👎

TomKimsey (approved):

oliver-walker-hms (approved):

alexjhawk:
Due to content changes by switching the deprecation back to `StringBuffer` from `StrBuilder`, I felt it was most appropriate to request reviews again.

oliver-walker-hms (approved):

TomKimsey (approved):
Commit message for 5c826c539569 could include details about the testing performed and that stringbuffer, although documented as the slowest, outperforms our stringbuilder and strBuilder.

alexjhawk:
@TomKimsey Commit message updated

TomKimsey (approved):

PR #33, opened 2022-12-16
Added local maven repo installation of library jar

TomKimsey:


alexjhawk:
How does this behave when you run the `package` phase multiple times with the same version number?

Things can get pretty messy with Maven if you overwrite or remove existing Maven repo files when they are being referenced by something else.

TomKimsey:
@alexjhawk @oliver-walker-hms not sure whats up with the code format check. Looks like the action cant find it.

TomKimsey:
> How does this behave when you run the `package` phase multiple times with the same version number?
> 
> Things can get pretty messy with Maven if you overwrite or remove existing Maven repo files when they are being referenced by something else.

Been working fine. Ive rewritten a version multiple times and the expected changes are present. For development its probably wise to use version numbers that are not expected to be used for real releases (ex "1.13.d")

alexjhawk (approved):

oliver-walker-hms (approved):

PR #35, opened 2022-12-16
Thk general changes

TomKimsey:
general changes that dont belong to a specific feature


src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppManagement.java
```diff
@@ -21,6 +21,17 @@ public class SCAppManagement {
   /** Seconds to pause thread when waiting for WAN IP */
   private static final int SECONDS_IN_WAIT_PERIOD = 20;
 
+  /**
```
TomKimsey, line 24: This is more personal preference and kind of a more c oriented behavior of mine. I prefer all the private methods and field at the top of the file.

oliver-walker-hms (approved):

alexjhawk (approved):

PR #38, opened 2022-12-20
Test pr

TomKimsey:
this is a test pull request


src/main/java/com/hms_networks/americas/sc/extensions/retry/AutomaticRetryResult.java
```diff
@@ -9,7 +9,7 @@
 public class AutomaticRetryResult {
 
   /** Integer assigned to represent a successful result. */
-  private static final int SUCCESS_INT = 0;
+  private static final int SUCCESS_INT = 5;
```
TomKimsey, line 12: this value should be 4.

TomKimsey, line 12:
```suggestion
  private static final int SUCCESS_INT = 4;
```

TomKimsey: @it-hms what do you mean by that?

PR #44, opened 2023-03-29
Dev/1.13.7

alexjhawk:


oliver-walker-hms (approved):

oliver-walker-hms:
There was an issue with github actions yesterday. I think they would pass if it was working
https://www.githubstatus.com/

alexjhawk:
> There was an issue with github actions yesterday. I think they would pass if it was working https://www.githubstatus.com/

It seems to be happening a lot, but thanks for letting me know. I want to wait until the actions pass because I've had issues with IntelliJ code formatting not catching everything since the latest update.

src/main/java/com/hms_networks/americas/sc/extensions/mqtt/ConstrainedMqttManager.java
```diff
@@ -665,6 +674,23 @@ public void setQos(int mqttQos) throws Exception {
     }
   }
 
+  /**
+   * Sets the time interval (in secs) for sending keep alive messages on the MQTT client connection.
+   * If the MQTT manager is currently running, the MQTT connection will be updated with the new keep
```
it-hms, line 679: Do you know if this should be set prior to start() being called? I think this comment might be misleading.  My guess is that when this option is set after the manager is already running there is no change i.e., this parameter is read when a connection is established.

it-hms (approved):
I approve, but I think the comment is misleading. Perhaps change the comment.

src/main/java/com/hms_networks/americas/sc/extensions/mqtt/ConstrainedMqttManager.java
alexjhawk: That is a good point. The Ewon MQTT implementation does not have developer documentation, and we only have a forum post to go off of: https://techforum.ewon.biz/thread-586.html

I traced the call down through the ETK MqttClient class, and it goes all the way into a native method, which we do not have visibility in to. 

It is probably best to assume that the MqttClient will not dynamically update the keep alive interval.

I will update this PR as follows:
1. Add (optional) constructor parameter for specifying keep alive interval.
2. Make `setMqttKeepAliveIntervalSecs` trigger an MQTT restart similar to the other parameters.

it-hms (approved):

alexjhawk:
@it-hms @oliver-walker-hms PR updated to improve the flow of setting the keep-alive interval. It can now be optionally set via the constructor (falls back to the default if not specified). 

I also added a commit to similarly improve the flow of configuring the MQTT thread sleep interval. It is now required to specify the MQTT thread sleep interval in the constructor. 
- Previously, it was necessary to call `setMqttThreadSleepIntervalMs()` before starting a connection. This could lead to unintended consequences (undefined sleep interval = 0) if the implementing developer forgot to make the call. This is no longer possible.

oliver-walker-hms (approved):

PR #49, opened 2023-05-31
Iatr/circular

### LINKED ISSUE #48: Circularized File exception check improve
Currently the Circularized file errors are checked and the default duration is 2 minutes. However this may cause repeated errors when the last read was successful -- this is because the check is always for two minutes . The check should be not for the last 2 minuets but since the last read try.

it-hms:
This update resolves a bug where the check for Circularized file looks back 1 minute in the Event File. This was changed to use absolute time and only look back to the most recent request. This error cased multiple error flags and requests. 

Closes #48 

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Ewon Flexy Extensions Library Changelog
 
+## Version 1.13.11
+### Features
+- N/A
+### Bug Fixes
+- Corrected check for circularized file error, to only check for error since the last EBD request. 
+
 ## Version 1.13.10
 ### Features
 - N/A
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
   <!-- PROJECT NAME -->
   <name>Ewon Flexy Extensions Library</name>
   <!-- PROJECT VERSION -->
-  <version>1.13.10</version>
+  <version>1.13.11</version>
   <!-- PROJECT GROUP ID (PARENT PACKAGE) -->
   <groupId>com.hms_networks.americas.sc</groupId>
   <!-- PROJECT ARTIFACT ID (ROOT PACKAGE NAME) -->
--- a/src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
@@ -7,6 +7,8 @@ import com.hms_networks.americas.sc.extensions.json.JSONException;
 import com.hms_networks.americas.sc.extensions.string.QuoteSafeStringTokenizer;
 import com.hms_networks.americas.sc.extensions.string.StringUtils;
 import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 import java.util.List;
 
 /**
@@ -21,6 +23,21 @@ public class EventFile {
 
   /** Event File line token delimiter */
   private static final String EVENT_FILE_LINE_TOKEN_DELIMITER = ";";
+
+  /** Base EBD request format for Event File (log) with absolute starting time */
+  private static final String EBD_EVENT_LOG_ABSOLUTE_START_BASE = "$dtEV$ftT$st";
+
+  /** Simple Date Format string for EBD date-time */
+  private static final String FLEXY_EBD_EVENT_FILE_DATE_FORMAT = "ddMMyyyy_HHmmss";
+
+  /** Event File expected number of tokens after splitting by delimiter */
+  private static final int EVENT_LOG_LINE_EXPECTED_TOKEN_SIZE = 6;
+
+  /** Event File line event ID offset */
+  private static final int EVENT_LOG_EVENT_ID_OFFSET = 5;
+
+  /** String encoding for reading Event File */
+  private static final String EVENT_FILE_STRING_ENCODING = "UTF-8";
 
   /**
    * Read EventFile, check for File Circularized event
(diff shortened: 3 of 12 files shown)


alexjhawk:
@it-hms Unless two applicable reviewers are unavailable (out of the office, etc.), a second reviewer should always be requested.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Ewon Flexy Extensions Library Changelog
```
alexjhawk, line 1: Release commit is missing required changes. See a previous release commit, such as https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/commit/550a530a65d67cfa96c51355b29d612dcff9b4d8.

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
```diff
@@ -123,4 +140,50 @@ private static boolean checkEventFileLine(String line, String target) {
...
+        HistoricalDataManager.executeEbdCall(
+            getEbdEventFileSinceAbsoluteTimeRequest(absoluteStartTimeMilliseconds));
+    String exporterFile = StringUtils.getStringFromInputStream(ex, EVENT_FILE_STRING_ENCODING);
+    ex.close();
+
+    if (exporterFile.length() > 0) {
+      final List eventFileLines = StringUtils.split(exporterFile, EVENT_FILE_LINE_DELIMITER);
+      for (int x = 1; x < eventFileLines.size(); x++) {
+        String line = (String) eventFileLines.get(x);
+        final List tokens = StringUtils.split(line, EVENT_FILE_LINE_TOKEN_DELIMITER);
```
alexjhawk, line 179: Is it possible that an event itself can contain the character `;`? If so, we should use `QuoteSafeStringTokenizer` here instead. An example of this can be found in the historical data manager: https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/blob/550a530a65d67cfa96c51355b29d612dcff9b4d8/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataManager.java#LL408C5-L408C30

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
it-hms: I think the only circumstance it if a use does the following in basic
`logevent "blah;blah;"`
However, I will change to safe tokenizer to handle this case

it-hms: @alexjhawk  @oliver-walker-hms  testing complete.

it-hms: @alexjhawk  @oliver-walker-hms  testing complete.

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
```diff
@@ -123,4 +140,45 @@ private static boolean checkEventFileLine(String line, String target) {
...
+    if (exporterFile.length() > 0) {
+      final List eventFileLines = StringUtils.split(exporterFile, EVENT_FILE_LINE_DELIMITER);
+      for (int x = 1; x < eventFileLines.size(); x++) {
+        String line = (String) eventFileLines.get(x);
+        if (checkEventFileLine(line, eventId)) {
+          return true;
+        }
+      }
+    }
+    return false;
```
alexjhawk, line 184:
```suggestion
    boolean eventFound = false;
    if (exporterFile.length() > 0) {
      final List eventFileLines = StringUtils.split(exporterFile, EVENT_FILE_LINE_DELIMITER);
      for (int x = 1; x < eventFileLines.size(); x++) {
        String line = (String) eventFileLines.get(x);
        if (checkEventFileLine(line, eventId)) {
          eventFound = true;
          break;
        }
      }
    }
    return eventFound;
```

Not sure how cleanly this suggestion will show, but we should avoid returning in the middle of a method where possible. It makes the method more difficult to follow because there is no longer only one entry and exit point.

I suggest we add a boolean such as `boolean eventFound = false` and then `return eventFound;` at the end of the method. If we find the event during the loop, then we can do `eventFound = false; break;`

src/main/java/com/hms_networks/americas/sc/extensions/package.html
```diff
@@ -5,7 +5,7 @@
 extension classes include many supplemental features, and the addition
 of new functionality which may not be present in the supported Java version.
 
-@version 1.13.10
+@version 1.14.10
```
alexjhawk, line 8:
```suggestion
@version 1.14.0
```
Version number incorrect

pom.xml
```diff
@@ -8,7 +8,7 @@
   <!-- PROJECT NAME -->
   <name>Ewon Flexy Extensions Library</name>
   <!-- PROJECT VERSION -->
-  <version>1.13.10</version>
+  <version>1.14.0</version>
```
alexjhawk, line 11: This is more of a discussion, but what warrants this version being 1.14.0 versus 1.13.11? 

A bump in the major version number seems inappropriate here, as there is only a single bug fix and no new features. (Per the changelog)

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
```diff
@@ -123,4 +140,45 @@ private static boolean checkEventFileLine(String line, String target) {
...
   }
+
+  /**
+   * Get the Export Block Descriptor request for Event File lines after the timestamp argument.
+   *
+   * @param timestampMilliseconds - the start time for request
+   * @return request string
+   */
+  private static String getEbdEventFileSinceAbsoluteTimeRequest(long timestampMilliseconds) {
+    Date TimeDateObj = new Date(timestampMilliseconds);
```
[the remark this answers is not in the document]
it-hms: will change

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
```diff
@@ -123,4 +140,45 @@ private static boolean checkEventFileLine(String line, String target) {
...
+   *
+   * @param eventId - ID of the target event
+   * @param absoluteStartTimeMilliseconds - epoch time in milliseconds as start time
+   * @return boolean - true/false found event ID in logs
+   * @throws EbdTimeoutException - for timeout of EBD request
+   * @throws IOException - for exception with input stream
+   */
+  public static boolean didEventOccurSinceAbsolute(
+      String eventId, long absoluteStartTimeMilliseconds) throws IOException, EbdTimeoutException {
+    Exporter ex =
```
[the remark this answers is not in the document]
it-hms: will change

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
it-hms: will change

CHANGELOG.md
alexjhawk, line 1: Typically the title for release commits would be in the following format:

`RELEASE: v1.13.11`

I just checked quickly and the title *shouldn't* affect any of our scripts/automation though.

alexjhawk (approved):
Looks good + ready after pending second approval from @oliver-walker-hms.

I only added one (optional) comment/suggestion.

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
```diff
@@ -16,6 +16,12 @@
  * @since 1.6.0
  */
 public class EventFile {
+  /** Event File line delimiter */
```
oliver-walker-hms, line 22: File does not need to be capitalized. This is repeated throughout the code on this and the next commit

oliver-walker-hms (approved):

src/main/java/com/hms_networks/americas/sc/extensions/eventfile/EventFile.java
alexjhawk: Agreed, although relatively minor

PR #51, opened 2023-06-12
Dev/1.14.1

alexjhawk:


src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -64,4 +64,87 @@ class HistoricalDataConstants {
...
+  public static final int QUEUE_DIAGNOSTIC_TAG_RUNNING_BEHIND_SECONDS_TYPE = 3;
+
+  /**
+   * The default threshold for the number of seconds that the queue is running behind before the
+   * diagnostic tag value is displayed. This value is used when one is not explicitly provided.
+   *
+   * <p>The running behind time display threshold helps to prevent users from being alarmed by a
+   * running behind time which is within acceptable limits.
+   */
+  public static final int DEFAULT_QUEUE_DIAGNOSTIC_TAG_RUNNING_BEHIND_SECONDS_DISPLAY_THRESHOLD = 3;
```
it-hms, line 96: Three seconds does not appear to be enough time to avoid false positives considering that we have configurable polling times. The default que poll time for Thingworx is 3 seconds, so any additional delays beyond this would indicate a false positive.  
 I would guess that this needs to be more like  20 or 30 seconds to avoid false positives.

it-hms (dismissed):
My only comment was about the default value for the 'is behind diagnostic tag'. I don't know if the default value is appropriate, or should be increased.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: I didn't realize that the check is after the EBD call. This makes the polling time less of an issue.

alexjhawk: I corrected this to `180`. The value of `3` was unrealistic, and I think I just made a copy/paste error or forgot to switch the value.

Realistically, connectors should be setting this threshold, so it shouldn't matter. For flexibility, I added a default value and the option to not specify though.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -64,4 +64,88 @@ class HistoricalDataConstants {
...
+   * has been polled for data.
+   */
+  public static final String QUEUE_DIAGNOSTIC_TAG_POLL_COUNT_DESC =
+      "Diagnostic tag containing the number of times the queue has been polled for data.";
+
+  /**
+   * The type of the diagnostic tag (integer) that is populated with the number of times that the
+   * queue has been polled for data.
+   */
+  public static final int QUEUE_DIAGNOSTIC_TAG_POLL_COUNT_TYPE = 3;
```
TomKimsey, line 147: For consts referencing tag types wouldnt it make sense to make the definitions in TagType.java public and use those. Also isnt 3 DWord?

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueDiagnosticTagManager.java
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * Attempts to create a diagnostic tag with the specified tag name, description, IO server, and
+   * type.
+   *
+   * @param tagName diagnostic tag name
+   * @param tagDesc diagnostic tag description
+   * @param tagIoServer diagnostic tag IO server
+   * @param tagType diagnostic tag type
+   * @return corresponding {@link TagControl} object for the diagnostic tag
+   * @throws Exception if the diagnostic tag is missing and cannot be created
```
TomKimsey, line 91: It can throw and EWException if its created but cannot be accessed via TagControl()

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Ewon Flexy Extensions Library Changelog
 
+## Version 1.14.1
+### Features
+- Added historical data queue diagnostic tag support to HistoricalDataQueueManager.java
```
TomKimsey, line 5:
```suggestion
- Added historical data queue diagnostic tag support to HistoricalDataQueueManager.java.
```

TomKimsey (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueDiagnosticTagManager.java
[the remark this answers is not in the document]
alexjhawk: Added similar

CHANGELOG.md
alexjhawk: Added

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataQueueDiagnosticTagManager.java
alexjhawk: Added

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
alexjhawk: Changed. And yes, I selected DWORD because it can handle larger values.

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -64,4 +66,89 @@ class HistoricalDataConstants {
...
+  /**
+   * The description of the diagnostic tag that is populated with the number of times that the queue
+   * has been polled for data.
+   */
+  public static final String QUEUE_DIAGNOSTIC_TAG_POLL_COUNT_DESC =
+      "Diagnostic tag containing the number of times the queue has been polled for data.";
+
+  /**
+   * The type of the diagnostic tag (integer) that is populated with the number of times that the
+   * queue has been polled for data.
```
TomKimsey, line 148: Should this comment be (DWORD) not (integer)?

alexjhawk: Yes

alexjhawk: Fixed

TomKimsey (approved):

it-hms (approved):

PR #59, opened 2023-09-15
V1.14.3

TomKimsey:


src/main/java/com/hms_networks/americas/sc/extensions/system/http/SCHttpUtility.java
```diff
@@ -11,7 +11,7 @@
  * Utility class for performing and managing HTTP connections on the Ewon Flexy system.
  *
  * @since 1.3.0
- * @version 1.0.0
+ * @version 1.0.1
```
TomKimsey, line 14: How are we handling version updates to files. I can fixup w/ previous but if there were multiple changes that required multiple commits prior to this I would expect this to be a seperate commit so im inclined to think its most proper in its own commit to indicate this is the state of v1.0.1.

alexjhawk: I've mainly been doing that to avoid frequently incrementing the class-wide @ version tag. That being said, for classes with an already established versioning scheme, I am ok with just sticking to that.

oliver-walker-hms (approved):

src/main/java/com/hms_networks/americas/sc/extensions/system/http/SCHttpUtility.java
TomKimsey: will fix

alexjhawk (approved):

oliver-walker-hms (approved):

src/main/java/com/hms_networks/americas/sc/extensions/system/http/SCHttpUtility.java
TomKimsey: Fixed

src/main/java/com/hms_networks/americas/sc/extensions/system/http/SCHttpUtility.java
```diff
@@ -48,6 +48,7 @@ public class SCHttpUtility {
    *
```
TomKimsey, line 51: Missed @ since on class level

src/main/java/com/hms_networks/americas/sc/extensions/system/http/SCHttpUtility.java
```diff
@@ -48,6 +48,7 @@ public class SCHttpUtility {
...
    * @param seconds the timeout in seconds
    * @throws Exception if unable to set timeouts
+   * @since 1.3.1
```
TomKimsey, line 54: these are wrong, should be 1.3.0. I have the fix ready to push

alexjhawk (approved):

oliver-walker-hms (approved):

PR #61, opened 2023-09-27
PR 2 for v1.15.0 (Changes Related to New Connector Development, Upstream Starter Project Changes, Etc.) 

alexjhawk:
This is PR 2 in a series of pull requests to bring in changes for v1.15.0. These are mostly related to the development of a new connector, staged changes that were previously untested, and other changes or enhancements, as described.

### DIFF
--- a/src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
@@ -1169,7 +1169,7 @@ public class JSONObject {
    *
    * @return a printable, displayable, portable, transmittable representation of the object,
    *     beginning with <code>{</code>&nbsp;<small>(left brace)</small> and ending with <code>}
-   * </code>&nbsp;<small>(right brace)</small>.
+   *     </code>&nbsp;<small>(right brace)</small>.
    */
   public String toString() {
     try {
@@ -1200,7 +1200,7 @@ public class JSONObject {
    * @param indentFactor The number of spaces to add to each level of indentation.
    * @return a printable, displayable, portable, transmittable representation of the object,
    *     beginning with <code>{</code>&nbsp;<small>(left brace)</small> and ending with <code>}
-   * </code>&nbsp;<small>(right brace)</small>.
+   *     </code>&nbsp;<small>(right brace)</small>.
    * @throws JSONException If the object contains an invalid number.
    */
   public String toString(int indentFactor) throws JSONException {
--- a/src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
@@ -24,6 +24,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
+import com.hms_networks.americas.sc.extensions.logging.Logger;
 import java.io.IOException;
 import java.io.Writer;
 import java.util.Enumeration;
@@ -647,6 +648,20 @@ public class JSONObject {
     return this;
   }
 
+  /**
+   * Put a non-null key/value pair in the JSONObject, where the value will be a non-null JSONArray
+   * that is produced from a Collection.
+   *
+   * @param key A key string (cannot be null).
+   * @param value A Collection value (cannot be null).
+   * @return this.
+   * @since 1.15.0
+   */
+  public JSONObject putNonNull(String key, Vector value) {
+    putNonNull(key, new com.hms_networks.americas.sc.extensions.json.JSONArray(value));
+    return this;
+  }
+
   // #if CLDC!="1.0"
 
   /**
@@ -809,6 +824,30 @@ public class JSONObject {
     return this;
   }
 
+  /**
+   * Put a non-null key/boolean pair in the JSONObject.
+   *
+   * <p>This method removes the {@link JSONException} thrown by {@link #put(String, boolean)} in
+   * favor of a cleaner solution (where applicable).
+   *
+   * <p>It is strongly recommended that the standard {@link #put(String, boolean)} method be used
+   * when any possibility of a null key exists, since it is entirely up to the calling method to
+   * ensure that the key is not null and the value is not invalid (if applicable).
+   *
+   * @param key A key string (cannot be null).
+   * @param value A boolean which is the value.
+   * @return this.
+   * @since 1.15.0
+   */
+  public JSONObject putNonNull(String key, boolean value) {
+    // #if CLDC!="1.0"
+    putNonNull(key, value ? Boolean.TRUE : Boolean.FALSE);
+    // #else
+    // #         put(key, value ? TRUE : FALSE);
+    // #endif
+    return this;
+  }
+
   // #if CLDC!="1.0"
 
   /**
@@ -821,6 +860,26 @@ public class JSONObject {
    */
   public JSONObject put(String key, double value) throws JSONException {
     put(key, new Double(value));
+    return this;
+  }
+
+  /**
+   * Put a non-null key/double pair in the JSONObject.
+   *
+   * <p>This method removes the {@link JSONException} thrown by {@link #put(String, double)} in
+   * favor of a cleaner solution (where applicable).
+   *
+   * <p>It is strongly recommended that the standard {@link #put(String, double)} method be used
+   * when any possibility of a null key or non-finite value exists, since it is entirely up to the
+   * calling method to ensure that the key is not null and the value is not invalid (if applicable).
+   *
+   * @param key A key string (cannot be null).
+   * @param value A double which is the value (cannot be NaN or infinite).
+   * @return this.
+   * @since 1.15.0
+   */
+  public JSONObject putNonNull(String key, double value) {
+    putNonNull(key, new Double(value));
     return this;
   }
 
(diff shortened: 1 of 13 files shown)


src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
```diff
@@ -809,6 +824,30 @@ public JSONObject put(String key, boolean value) throws JSONException {
...
+   * @param key A key string (cannot be null).
+   * @param value A boolean which is the value.
+   * @return this.
+   * @since 1.15.0
+   */
+  public JSONObject putNonNull(String key, boolean value) {
+    // #if CLDC!="1.0"
+    putNonNull(key, value ? Boolean.TRUE : Boolean.FALSE);
+    // #else
+    // #         put(key, value ? TRUE : FALSE);
```
oliver-walker-hms, line 846: the spacing looks odd here. I've seen other preprocessor directives in java where there is no space before the hashtag, does this still work?

Also, the spacing before put looks odd.

alexjhawk: I'm hesitant to change this formatting, as it matches the already established formatting for directives throughout the file. This file was adopted from public source JSON classes/files.

oliver-walker-hms: sounds like it probably works as is.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
```diff
@@ -311,6 +311,34 @@ public static void LOG_DEBUG(String logString) {
     LOG(LOG_LEVEL_DEBUG, logString);
   }
 
+  /**
+   * Log a string and exception with the debug log level
```
it-hms, line 315: Perhaps is top-level comment is a sentence, it should end in period.  Not sure if this looks better or worse for doc strings. 
For params and returns, this is not necessary.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
```diff
@@ -320,6 +348,33 @@ public static void LOG_INFO(String logString) {
...
+  }
+
+  /**
+   * Log a string and exception with the info log level if {@code elevateExceptionLogLevel} is
+   * {@code true}, otherwise log the exception with the trace log level.
+   *
+   * @param logString string to log
+   * @param logException exception to log
+   * @param elevateExceptionLogLevel elevates the log level used for the specified exception to
+   *     debug
```
it-hms, line 369: to info?

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
```diff
@@ -329,6 +384,33 @@ public static void LOG_WARN(String logString) {
...
+  }
+
+  /**
+   * Log a string and exception with the warning log level if {@code elevateExceptionLogLevel} is
+   * {@code true}, otherwise log the exception with the trace log level.
+   *
+   * @param logString string to log
+   * @param logException exception to log
+   * @param elevateExceptionLogLevel elevates the log level used for the specified exception to
+   *     debug
```
it-hms, line 405: to warn?

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
```diff
@@ -338,6 +420,33 @@ public static void LOG_SERIOUS(String logString) {
...
+  }
+
+  /**
+   * Log a string and exception with the serious log level if {@code elevateExceptionLogLevel} is
+   * {@code true}, otherwise log the exception with the trace log level.
+   *
+   * @param logString string to log
+   * @param logException exception to log
+   * @param elevateExceptionLogLevel elevates the log level used for the specified exception to
+   *     debug
```
it-hms, line 441: to serious?

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
```diff
@@ -347,6 +456,33 @@ public static void LOG_CRITICAL(String logString) {
...
+  }
+
+  /**
+   * Log a string and exception with the critical log level if {@code elevateExceptionLogLevel} is
+   * {@code true}, otherwise log the exception with the trace log level.
+   *
+   * @param logString string to log
+   * @param logException exception to log
+   * @param elevateExceptionLogLevel elevates the log level used for the specified exception to
+   *     debug
```
it-hms, line 477: to critical?

starting-files/jvmrun
```diff
@@ -0,0 +1 @@
+-heapsize 25M -classpath /usr/extensions-1.14.3-full.jar -emain com.hms_networks.americas.sc.extensions.ExtensionsMain
```
it-hms, line 1: This doesn't really matter, but this is a pr for 1.15, so won't this get updated in a couple commits?

it-hms (changes requested):
I think I only had issues with some doc strings.

TomKimsey (commented):
I dont have anything to add

starting-files/jvmrun
alexjhawk: Yes. To ensure consistency with release commits across all starter project-based code bases, I added this as 1.14.3. That way, when a release commit is merged for v1.15.0 in a future PR, the version number will still be updated in all expected locations.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: Will make these changes shortly.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: Added punctuation. I initially did not include these because it maintained consistency with the class, but I have added an additional commit to address this.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: Updated

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: Updated

src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
alexjhawk: This whitespace is consistent with the existing conventions in this file as-is. 

Additional changes to the formatting here should be regarded as out-of-scope, and any concerns may be better addressed in a separate PR/issue.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: In the future, what do you think about documenting consistency concerns like this and then circling back after the PR?

- It makes sense to address the punctuation directly added/modified/lack thereof in the pull request, but subsequent concerns about consistency could be more effectively addressed.
- In the past, we've mostly addressed the consistency concern(s) in the PR, but this also introduces scope creep to PRs, making them larger and more cumbersome to review and modify.
- Scope creep in PRs can also increase the time required to work with a PR because it can introduce unexpected modification/rebasing of content that was otherwise edited in other commits and could conflict with other staged commits in certain scenarios, depending on order.

starting-files/jvmrun
TomKimsey: Would it make sense to have this be introduced as 
```
-heapsize 25M -classpath /usr/extensions-0.0.0-full.jar -emain com.hms_networks.americas.sc.extensions.ExtensionsMain
```
To make it clear that this is introduced as unusable until the next release?

alexjhawk: The `pom.xml` at this stage still contains `1.14.3` as the version number, so I think this is fine as-is. It is basically dropping in files that would already be expected here per our starter project structure/conventions.
- There is no technical reason that this would not function with v1.14.3, and having v0.0.0 would make the file more unusable since v0.0.0 doesn't exist.

Typically, we would expect this file to be present during the `RELEASE: X.Y.Z` commit stage of a release, so I introduced it with the version number that would be expected at this point in the Git history (given the pom.xml, etc.). This also allows us to change the value from `v1.14.3` to `v.1.15.0` as expected during the future release commit.

alexjhawk: Also, in other repositories, we have merged parts of a version to be released and maintained the existing values in all applicable locations, including the `jvmrun` file. The only difference here is that this file did not exist previously, because this repository was not entirely in sync with the upstream starter project.

alexjhawk: > That was the intent

If you and @it-hms both prefer to use `v0.0.0`, then I will make this change, but I think it makes the overall Git history messier because, at this point in the repository, one would expect this file to be at `v1.14.3`. 

`v0.0.0` seems arbitrary to me, is not a valid version number for this repository, and is not applicable to the current state of the repository.

TomKimsey: im fine with it as is. Im just offering an alternative.

alexjhawk: Ok. I'll leave it as-is.

src/main/java/com/hms_networks/americas/sc/extensions/json/JSONObject.java
oliver-walker-hms: I've added a new [issue](https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/issues/63) to look into this later.
It's fine as is for now.

it-hms (approved):

TomKimsey (approved):

oliver-walker-hms (approved):

alexjhawk:
I expect to have PR 3 ready either later today or tomorrow. It should be the final PR and will mainly consist of web docs (how to use the library only, no detailed documentation...yet.)

PR #72, opened 2023-10-23
Thk watchdog dev

TomKimsey:
The reboot watchdog functionality currently can be incorrectly used by the application if it implements servicing too quickly. This addition adds the functionality to enable a timer. The timer ensures that a specified amount of time passes before the watchdog is serviced. Ideally this would be in the 5-10 minute range (long enough to establish a VPN connection with the device). 

This prevents situations where the device is cyclically rebooting once every two minutes and the reboot watchdog is not triggered but a VPN connection is not possible.,

### DIFF
--- a/src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
@@ -3,6 +3,7 @@ package com.hms_networks.americas.sc.extensions.system.application;
 import com.hms_networks.americas.sc.extensions.config.ConfigFileAccessManager;
 import com.hms_networks.americas.sc.extensions.json.JSONException;
 import com.hms_networks.americas.sc.extensions.json.JSONObject;
+import com.hms_networks.americas.sc.extensions.system.time.SCTimeUnit;
 import java.io.File;
 import java.io.IOException;
 
@@ -40,6 +41,9 @@ public class SCAppRebootWatchdog {
   /** Max number of reboots before watchdog disables the jvmrun file */
   private static final int MAX_REBOOT_COUNT = 10;
 
+  /** Default time the watchdog has to run before it can be serviced */
+  private static final int DEFAULT_WATCHDOG_MIN_SERVICE_TIME_SEC = 0;
+
   /** Contents of reboot watchdog data file */
   private static JSONObject rebootWatchdogObject;
 
@@ -49,13 +53,37 @@ public class SCAppRebootWatchdog {
   /** Status indicator if reboot watchdog is serviced */
   private static boolean isRebootWatchdogServiced = false;
 
+  /** The time the watchdog will accept servicing */
+  private static long watchdogReadyTimeMillis;
+
   /**
    * Enables the reboot watchdog.
+   *
+   * <p>The watchdog will be enabled with a default service time of 0 seconds. This means that when
+   * using the default service time, the watchdog can be serviced immediately after initialization.
+   * This ensures compatibility with existing applications that have implemented the reboot
+   * watchdog, but have not incorporated a sizeable delay between watchdog initialization and
+   * servicing.
+   *
+   * <p>For new applications, this method should not be used. Instead, the {@link
+   * #enableRebootWatchdog(int)} method should be used.
    *
    * @throws IOException If watchdog file cannot be read from filesystem.
    * @throws JSONException If watchdog file is not valid JSON or doesn't have the correct structure.
    */
   public static void enableRebootWatchdog() throws IOException, JSONException {
+    enableRebootWatchdog(DEFAULT_WATCHDOG_MIN_SERVICE_TIME_SEC);
+  }
+
+  /**
+   * Enables the reboot watchdog.
+   *
+   * @param serviceTimeSec The time the watchdog has to run before it can be serviced
+   * @throws IOException If watchdog file cannot be read from filesystem.
+   * @throws JSONException If watchdog file is not valid JSON or doesn't have the correct structure.
+   */
+  public static void enableRebootWatchdog(int serviceTimeSec) throws IOException, JSONException {
+    setWatchdogReadyTime(SCTimeUnit.SECONDS.toMillis(serviceTimeSec));
     initRebootWatchdogFile();
     isRebootWatchdogInit = true;
   }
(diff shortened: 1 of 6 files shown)


src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
```diff
@@ -190,4 +221,25 @@ private static JSONObject getDefaultRebootWatchdogObject() throws JSONException
...
+
+  /**
+   * Checks if the watchdog has been running for a sufficient amount of time
+   *
+   * @return A boolean indicating if the watchdog has been running for a sufficient amount of time
+   */
+  private static boolean checkWatchdogWaitTimer() {
+    long currentTimeMillis = System.currentTimeMillis();
+
+    return (currentTimeMillis >= watchdogReadyTimeMillis) ? true : false;
```
alexjhawk, line 243:
```suggestion
    return currentTimeMillis >= watchdogReadyTimeMillis;
```

`? true : false;` is redundant.

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
TomKimsey: fixed.

alexjhawk (approved):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
alexjhawk: I agree. I only caught this because IntelliJ shows a warning when you use `? true : false` so it is engraved in my mind always to check.

alexjhawk (changes requested):
I just noticed that the _pom_library.mdx file was not also changed in the release commit. 

Previous release commit: https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/commit/837dec3bd9e577dd55ea9f00df933f967c9f0caf

Some of this ultimately should be automated. I don't see any reason why we couldn't make a JS component that queries GitHub Releases to get the latest release version number. It could also be specially added in to the Maven config to replace the value automatically (similar to current setup for IDE debug configs and jvmrun file automatic updates buried in the pom.xml file)

alexjhawk (dismissed):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
```diff
@@ -190,4 +221,25 @@ private static JSONObject getDefaultRebootWatchdogObject() throws JSONException
...
+
+  /**
+   * Checks if the watchdog has been running for a sufficient amount of time
+   *
+   * @return A boolean indicating if the watchdog has been running for a sufficient amount of time
+   */
+  private static boolean checkWatchdogWaitTimer() {
+    long currentTimeMillis = System.currentTimeMillis();
+
+    return currentTimeMillis >= watchdogReadyTimeMillis;
```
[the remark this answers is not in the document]
TomKimsey: changed, although im indifferent on it.

alexjhawk (approved):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
```diff
@@ -49,13 +53,28 @@ public class SCAppRebootWatchdog {
   /** Status indicator if reboot watchdog is serviced */
   private static boolean isRebootWatchdogServiced = false;
 
+  /** The time the watchdog will accept servicing */
+  private static long watchdogReadyTimeMillis;
+
   /**
-   * Enables the reboot watchdog.
+   * Enables the reboot watchdog. The watchdog will be enabled with a default service time of 0.
```
it-hms, line 60: Could you please add a comment about the default value of 0. What does 0 mean? Does this mean the watchdog will be serviced immediately after boot? Instead of waiting for after stability?

it-hms (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
TomKimsey: added

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
```diff
@@ -49,13 +53,33 @@ public class SCAppRebootWatchdog {
...
+  private static long watchdogReadyTimeMillis;
+
   /**
-   * Enables the reboot watchdog.
+   * Enables the reboot watchdog. The watchdog will be enabled with a default service time of 0
+   * seconds. This means that when using the default watchdog can be serviced immediately after
+   * initialization. This ensures compatibility with existing applications that have implemented the
+   * reboot watchdog but have not incorporated a sizeable delay between watchdog initialization and
+   * servicing. For new applications this method should not be used and instead the method with a
+   * service time parameter should be used.
```
alexjhawk, line 65:
```suggestion
   * Enables the reboot watchdog. 
   * <p>
   * The watchdog will be enabled with a default service time of 0 seconds. This means that when
   * using the default service time, the watchdog can be serviced immediately after initialization. 
   * This ensures compatibility with existing applications that have implemented the reboot
   * watchdog, but have not incorporated a sizeable delay between watchdog initialization and
   * servicing.
   * <p>
   * For new applications, this method should not be used. Instead, the 
   * {@link #enableRebootWatchdog(int)} method should be used.
```

alexjhawk: Grammar and spacing changes for readability.

alexjhawk (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/system/application/SCAppRebootWatchdog.java
TomKimsey: changed

alexjhawk (approved):

it-hms (approved):

PR #77, opened 2023-11-01
Thk v1.15.6

TomKimsey:


.github/workflows/package-test.yml
```diff
@@ -1,12 +1,12 @@
...
-# Compile Test Action for Maven-based Ewon ETK Projects
+# Package Test Action for Maven-based Ewon ETK Projects
 # Version: 2.1
 #
 # This action is configured to automatically run when a push
 # is made or pull request is merged to the `main` or `master`
 # branch.
 
-name: Compile Test
+name: Package Test
```
alexjhawk, line 9: What do you think about `Package Build Test` or `Packaging Test`? I don't think it's a huge concern, but `Package Test` just seems off to me.

TomKimsey: This is still in draft. not sure why i added a reviewer yet.

Im fine with "Package Test" or "Build Test", "Package Test" doesnt bother me at all.

pom.xml
```diff
@@ -171,6 +171,7 @@
           <dependencySourceExcludes>
             <dependencySourceExclude>com.hms_networks.americas.sc.mvnlibs:ewon-etk:*</dependencySourceExclude>
           </dependencySourceExcludes>
+          <failOnWarnings>true</failOnWarnings>
```
alexjhawk, line 174: This should be carried upstream to the starter project.

.github/workflows/package-test.yml
alexjhawk: I'm okay with either, although `Build Test` may be a bit more descriptive as it covers the Javadoc checking too.

alexjhawk (dismissed):

pom.xml
TomKimsey: Yeah I agree

alexjhawk (approved):

PR #81, opened 2023-11-02
Dev/1.15.7

alexjhawk:


src/main/java/com/hms_networks/americas/sc/extensions/datapoint/DataPoint.java
```diff
@@ -133,9 +133,16 @@ public String toString() {
   public abstract DataType getType();
 
   /**
-   * Get the string representation of the data point value.
+   * Get the {@link String} representation of the data point value.
```
TomKimsey, line 136: This belongs in a separate commit

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
```diff
@@ -0,0 +1,312 @@
...
+   *
+   * @param values values to sum
+   * @return sum of values
+   * @throws IllegalArgumentException if values are not of the same type
+   * @since 1.15.7
+   */
+  public static Object getValueSum(Object[] values) {
+    // Sum values by type
+    Object sum;
+    if (values[0] instanceof Integer) {
```
TomKimsey, line 209: 0 should be a named variable.

TomKimsey (changes requested):

src/main/java/com/hms_networks/americas/sc/extensions/datapoint/DataPoint.java
alexjhawk: Changed (and covered the instances of this in subclasses)

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: Moved the entire `values[0]` out to a variable because it is referenced multiple times below.

TomKimsey (dismissed):

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
```diff
@@ -0,0 +1,313 @@
...
+    return minValue;
+  }
+
+  /**
+   * Get and return the maximum value of the two specified values.
+   *
+   * @param value1 first value
+   * @param value2 second value
+   * @return maximum value
+   * @throws IllegalArgumentException if values are not of the same type
```
it-hms, line 106: or are not types that support comparison

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
```diff
@@ -0,0 +1,313 @@
...
+   * @return sum of values
+   * @throws IllegalArgumentException if values are not of the same type
+   * @since 1.15.7
+   */
+  public static Object getValueAvg(Object[] values) {
+    // Get sum of values
+    Object sum = getValueSum(values);
+
+    // Get average of values
+    Object avg = getValueDivide(sum, values.length);
```
it-hms, line 192: for empty array, divide by zero. Perhaps check and throw nicer exception.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
```diff
@@ -0,0 +1,313 @@
...
+    // Return average
+    return avg;
+  }
+
+  /**
+   * Get and return the sum of the specified values.
+   *
+   * @param values values to sum
+   * @return sum of values
+   * @throws IllegalArgumentException if values are not of the same type
```
it-hms, line 203: I don't think there is any checking if values are the same type.  
If values are not types that support addition? 
Should add comment that caller is responsible values being a homogenous list

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
```diff
@@ -0,0 +1,313 @@
...
+  /**
+   * Get and return the value of the specified value divided by the specified divisor.
+   *
+   * @param value the value to divide
+   * @param divisor the divisor
+   * @return the value divided by the divisor
+   * @throws IllegalArgumentException if the value is not of a supported type
+   * @since 1.15.7
+   */
+  public static Object getValueDivide(Object value, int divisor) {
```
it-hms, line 271: divisor is an int, so this is meant for average calculation, not really general purpose. Should be private? Or should  divisor be object type?

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
```diff
@@ -0,0 +1,313 @@
...
+    } else if (firstValue instanceof Short) {
+      // Sum values
+      short sumShort = 0;
+      for (int i = 0; i < values.length; i++) {
+        sumShort += ((Short) values[i]).shortValue();
+      }
+      sum = new Short(sumShort);
+    } else if (firstValue instanceof Byte) {
+      // Sum values
+      byte sumByte = 0;
```
it-hms, line 247: will there be a rollover if sum of bytes>255?  Perhaps sum should always return long or double?

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
```diff
@@ -0,0 +1,313 @@
...
+
+  /**
+   * Get and return the sum of the specified values.
+   *
+   * @param values values to sum
+   * @return sum of values
+   * @throws IllegalArgumentException if values are not of the same type
+   * @since 1.15.7
+   */
+  public static Object getValueAvg(Object[] values) {
```
it-hms, line 187: This implementation does not work well for Bytes. If byte[] is longer than 255 or sum > 255

it-hms (changes requested):
I don't think getValueAvg will work for Bytes due to roll over. May not work for ints too, but this is less of a concern.  Perhaps needs to throw rollover exception, or needs to perform sum and division with long types.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: This is a side effect of supporting all of the primitive types in Java, and this class would likely not be necessary on Java 1.5+.

It doesn't make sense for us to add special handling based on specific types. Ultimately, this is a helper method that 'replaces' newer Java code like this:

```
int sum = 0;
for (int x: values) sum += x;
```

Where `sum += x` is the same as `sum = sum + x;`, the same issue could arise here regardless of type. It is a general consideration when summing numbers, and our implementation should handle this exactly the same as using a loop to sum the values in newer Java, without a utility class.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: Java has underflow/overflow, but it does not explicitly catch or throw an error in that event.

This implementation matches what is normally expected out of newer Java code, where the sum of two integers would return an integer. The JVM does not automatically convert types to work around overflow concerns. For example, in the following Java code, the raw type `T` would be `int` (note: Java does not actually allow raw/parameterized primitive types, this is just an example):

```
int x = 10;
int y = 10;
T sum = x + y;
```

Regardless of the values in `x` and `y`, the type for `sum`, `T` is always going to be `int`. If the integers are too large, it would be expected that they are converted *beforehand* (before being summed, or in this case, passed to the util method). This would look like 

```
int x = 10;
int y = 10;
T sum = (long) x + (long) y;
```

where `T` would in this case be the type `long`.

By changing the return type to be something other than the input type(s), the calling method is likely to run into errors as well, because it would expect the result to be of the same type passed in, and then a class cast exception would occur when trying to convert the returned `Object` to the expected type, because the type does not match.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: I will add a check. Given the constraints imposed by `getValueSum(...)`, an empty array would cause an error before reaching here, but it doesn't hurt to double-check.

It would be expected that in `getValueSum(...)` since the first value is null, it would not meet any of the conditions such as `instanceof Integer`, `instanceof Double`, etc, thus resulting in an IllegalArgumentException with a message similar to: `Cannot sum values of type null.`

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
it-hms: What do you think is a good solution? drop support for Bytes?

TomKimsey: As an alternative, what about creating a OverflowException?

alexjhawk: I don't think any changes are required. There is no reason for us to deviate from what the JVM would normally do, and this method is no different from something like
`int a = b + c;` except we are casting the types based on `instanceof` due to Java 1.4 limitations.

Regarding support for 'Byte', this likely won't receive much use but is included so that the utility class maintains support for *all* primitive types.

alexjhawk: > As an alternative, what about creating a OverflowException?

This could be done, but at the end of the day, dealing with overflow/underflow is out of the scope here I think.

This method is concerned with casting the types (as detected) and then summing the two number values exactly as they would normally in Java, with no extra processing or handling.

TomKimsey: Im not sure its fair to compare an operator to a method. Using operators you always have the ability to cast. In this method you dont provide that ability. IF you could specify the return type(/type to use for summation) i think that would be more comparable.

alexjhawk:
I am reviewing two customer support requests and will revisit the outstanding changes on this PR later today.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: The intention of this method is to act as an operator, though, with the advantage being that it hides a lot of the casting required in Java 1.4 when using primitive values in their corresponding `Object` class, i.e. `Integer`. This includes the side effects, positive or negative, from the values you put in.

I will take a further look at how the JVM handles summation with multiple values and types. If there are improvements that can be made in this method to more closely mimic the expectations/actual JVM behavior (when directly summing), I will change.

The goal here is to be able to have two or more data points from the same tag, take their values from `DataPoint#getValueObject()`, and then get the corresponding min/max/avg values without the large amounts of type detection and casting code at the connector/application level.

alexjhawk:
> I am reviewing two customer support requests and will revisit the outstanding changes on this PR later today.

I am back to working on this PR. I have made a few improvements to the divide method, and have a few other changes upcoming which will better align with the exact JVM behaviors. Alongside this will be a JUnit test class to verify the functionality of these methods.

alexjhawk:
> > I am reviewing two customer support requests and will revisit the outstanding changes on this PR later today.
> 
> I am back to working on this PR. I have made a few improvements to the divide method, and have a few other changes upcoming which will better align with the exact JVM behaviors. Alongside this will be a JUnit test class to verify the functionality of these methods.

I have made several changes to better align with the exact JVM behavior if the operations were performed directly with the primitive values. The JUnit tests I added to go along with this class are complete and fairly extensive, given the assistance of GitHub CoPilot. Unfortunately, while many of these tests are passing, a few fail, so I need to address those discrepancies first.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
```diff
@@ -0,0 +1,313 @@
...
+ */
+public class RawNumberValueUtils {
+
+  /**
+   * Get and return the minimum value of the two specified values.
+   *
+   * @param value1 first value
+   * @param value2 second value
+   * @return minimum value
+   * @throws IllegalArgumentException if values are not of the same type
```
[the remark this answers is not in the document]
alexjhawk: I've made numerous changes to the Javadocs to more clearly outline the input restrictions and the resulting exceptions should those be violated.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/pull/81#discussion_r1385311719

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: https://github.com/hms-networks/sc-ewon-flexy-extensions-lib/pull/81#discussion_r1385311719

`Values` does not necessarily need to be a list of all the same types, as long as they are all implementations of `Number`. Checking has been added/ensured for this.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: Updated so `divisor` is an `Object`.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: A method overload was added to allow for overflow handling of the integer type. This covers byte, short, and int. If an overflow is detected and the option is true, the returned result is promoted to `Long` so it contains the true result. This is in addition to supporting the primitive operation behavior where the overflow would still return an integer result.

src/main/java/com/hms_networks/americas/sc/extensions/util/RawNumberValueUtils.java
alexjhawk: Java actually does do promotion from 'byte' or 'short' to an 'int' when performing operations on the values. I have updated all applicable methods/code to account for this. The JUnit test class also verifies this behavior.

alexjhawk:
All JUnit test failures have been resolved and PR is ready for review again. 

A majority of the failures were associated with the improper handling of short/byte operations, and the remaining few were resolved by correcting tests to account for the operation results for byte/short types to be int.

Basically, Java does not handle overflow for the corresponding primitive operations, but it does promote byte/short to int when performing operations, which resolves overflow issues/concerns for those two types. This has been changed/reflected in the code now, and JUnit tests have been added to verify compliance with all of the supported types.

it-hms (dismissed):

web-docs/docs/02-CHANGELOG.mdx
```diff
@@ -23,7 +32,8 @@ toc_max_heading_level: 2
 
 ## Version 1.15.3
 ### Features
-- Added service timer to the application reboot watchdog to ensure application is in a stable state before allowing the watchdog to be serviced.
+- Added service timer to the application reboot watchdog to ensure application is in a stable state
```
TomKimsey, line 35: A lot of previous versions are getting touched here. These do not belong in this commit. Also is there the intention to add a max line length parameter to .mdx files?

TomKimsey (changes requested):

web-docs/docs/02-CHANGELOG.mdx
alexjhawk: I will correct this. I didn't intend to establish a maximum length in MDX files, although IntelliJ formats according to the same max line length as our Java code. I just missed the changes leaking into the release commit.

That being said, any new MD/MDX I create generally does adhere to the same max line length requirement as our Java code.

alexjhawk: Fixed

TomKimsey (approved):

PR #82, opened 2023-11-07
Updated Changelog

TomKimsey:
@alexjhawk I often want to make changes that dont warrant a new release. How do you feel about adding commits like this where the Version is not set in the changelog but the entry is started. I just want a way to keep up on the changelog and allow multiple devs to potentially contribute to a version release with the changelog being accurate.

DO NOT MERGE

### DIFF
--- a/web-docs/docs/02-CHANGELOG.mdx
+++ b/web-docs/docs/02-CHANGELOG.mdx
@@ -5,6 +5,11 @@ sidebar_label: Change Log
 toc_max_heading_level: 2
 ---
 
+## Version CHANGEME
+### Features
+- Feature 1
+- Feature 2
+
 ## Version 1.15.6
 ### Features
 - Added missing Javadocs


alexjhawk:
I like this concept, although we may be able to find a better filler for `CHANGEME`. It doesn't really matter what we use at the end of the day, because it should never be merged, therefore it should never affect the automation. 

That being said, something like `v1.15.6-next1` or `v1.15.7-pre1` may fit better. We could stick with the existing version number and then use `nextX` to indicate active development of the next version on top of that, or we could bump the version number as appropriate and just include `preX`.

A more conventional approach would probably be to use a `preX` pattern. If a change is introduced that warrants a larger version number bump, I think that would be fine. I don't really have issues with 'pre' versions that don't have an associated official release, as long as the contained changes are in an 'official' release with the same/higher version number than the version number associated with the 'pre' build. I've seen plenty of examples of this in other Maven projects.

TomKimsey:
> I like this concept, although we may be able to find a better filler for `CHANGEME`. It doesn't really matter what we use at the end of the day, because it should never be merged, therefore it should never affect the automation.
> 
> That being said, something like `v1.15.6-next1` or `v1.15.7-pre1` may fit better. We could stick with the existing version number and then use `nextX` to indicate active development of the next version on top of that, or we could bump the version number as appropriate and just include `preX`.
> 
> A more conventional approach would probably be to use a `preX` pattern. If a change is introduced that warrants a larger version number bump, I think that would be fine. I don't really have issues with 'pre' versions that don't have an associated official release, as long as the contained changes are in an 'official' release with the same/higher version number than the version number associated with the 'pre' build. I've seen plenty of examples of this in other Maven projects.

I would like to avoid anticipating the version number since multiple developers could be working on changes at the same time and a new version is published before the change set that is updating the changelog gets approved. 

I would be okay with something like vX.Y.Z as its pretty evident its not a version and it needs changing. Or vM.m.p , vMajor.Minor.Patch, v#.#.#

PR #85, opened 2023-12-13
Dev/1.15.9

alexjhawk:


src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
```diff
@@ -80,6 +82,17 @@ public class Logger {
...
+  /**
+   * List for tracking whether a specific log has been outputted. This is used to prevent spamming
+   * the log with the same message.
+   *
+   * <p>If the log has already been outputted, its key will be in the list, otherwise it will not be
+   * present.
+   *
+   * @since 1.15.9
+   */
+  private static final List logOnceList = new ArrayList(); // List<String>
```
TomKimsey, line 94: Should this be implemented as a FIFO instead to protect against this list getting too large?

alexjhawk:
To make access easier to 'log once' functionality, I have enabled it directly within the `Logger.java` class. This has also been enabled within the configuration class in the abstract connector framework.

src/main/java/com/hms_networks/americas/sc/extensions/connectors/framework/AbstractConnectorConfig.java
```diff
@@ -538,12 +538,32 @@ public JSONObject getDefaultConfigurationObject() throws JSONException {
...
+  protected void logMissingField(String missingKey, String defaultValue, boolean logOnce) {
+    String logMessage =
         "The "
             + missingKey
             + " option was not found in the configuration file. Using default value of "
             + defaultValue
-            + ".");
+            + ".";
+    if (logOnce) {
+      Logger.LOG_WARN_ONCE(missingKey, logMessage);
```
TomKimsey, line 563: this key implementation is not very unique. If another log message was added referencing the same config file key and used the same scheme it would not be printed but it would be unique.

TomKimsey:
Or you could just hash the log message string to store in a limited size arraylist. Then just compare hashes.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: I don't know what you mean by using a FIFO here. 

This was implemented with the reduction of memory in mind, thus the reason for storing *only* the key and not the body of the log message.

alexjhawk:
> Or you could just hash the log message string to store in a limited size arraylist. Then just compare hashes.

I considered this solution, but I think it is just extra processing that is ultimately unnecessary. 

Also, using the log once methods in the logger is the responsibility of the implementing application, and I don't think it makes sense to spend additional runtime power in the library to work around the potential for compiled code that contains improper usage.
- If an application uses the same key twice currently, the worst-case scenario is that a log message is not output.
- In the proposed solution, the worst-case scenario would be much more impactful but also unknown. The implementing application invokes the log once method(s) an unknown number of times with messages that are an unknown size, and calculating hashes has the potential to drag down performance in connectors.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
TomKimsey: If an application were to use this functionality in a way that is not in the spirt of the implementation it may generate up to an infinite amount of new keys over a a long enough runtime. 

Example: (this would be bound by the amount of values possible for whatever datatype tag is.
`Logger.LOG_WARN_ONCE(tag.valuestring(), tagValueOutOfBounds);`

if it was implemented as a fifo you could allow the logOnce list to grow to a set amount of entries, then trim the oldest so that even if the mechanism was abused you would not have memory issues.

src/main/java/com/hms_networks/americas/sc/extensions/connectors/framework/AbstractConnectorConfig.java
alexjhawk: This method is within the abstract connector framework's configuration class, and the goal of the `.logMissingField` methods is to provide a single point for handling log output associated with missing configuration fields. 

I opted for directly using the key here because it is generally safe to assume there is no other need for another/other log messages elsewhere.

TomKimsey:
> > Or you could just hash the log message string to store in a limited size arraylist. Then just compare hashes.
> 
> I considered this solution, but I think it is just extra processing that is ultimately unnecessary.
> 
> Also, using the log once methods in the logger is the responsibility of the implementing application, and I don't think it makes sense to spend additional runtime power in the library to work around the potential for compiled code that contains improper usage.

**The javadocs do not suggest or define "proper usage".** 

> 
> * If an application uses the same key twice currently, the worst-case scenario is that a log message is not output.

**-Not printing a message because of poor key choice would be bad.**

> * In the proposed solution, the worst-case scenario would be much more impactful but also unknown. The implementing application invokes the log once method(s) an unknown number of times with messages that are an unknown size, and calculating hashes has the potential to drag down performance in connectors.

The javadocs do not state that this should not be used in cyclic operations and the functionality suggests thats exactly how it should be used. Under "normal" conditions the performance impact would probably not be measurable.

src/main/java/com/hms_networks/americas/sc/extensions/connectors/framework/AbstractConnectorConfig.java
TomKimsey: I think even a simple improvement would be 

`Logger.LOG_WARN_ONCE("logMissingField:"+missingKey, logMessage);`

Its not unreasonable to thing someone may add in somewhere else
`Logger.LOG_WARN_ONCE(invalidValueKey, logMessage);`
and it would not be printed because the key could be exactly the same.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: This could be implemented, but it was outside my original scope and makes the implementation less simple/straightforward.

This also checks for/guards against a case that I think is an application-level concern. We don't currently impose limits on other ArrayLists, including those used for retry messages, queued payloads, response error messages, etc., in our connectors, most of which are certain to be larger and more volatile.

TomKimsey: This is a public library and as such i think we should aim to provide the most stable library solution we can for users. I dont expect users of the library to dig into the source to attempt to find potential issues/bugs. 

I do not think it would be unreasonable to have a sane default, then allow for limit adjustment via a method
`Logger.setOnceMessageListSize(100)`
`Logger.setOnceMEssageListSize(-1) //no limit`

We probably should look into other uses of arraylist and see if we can improve the user experience in terms of noting in the javadocs implications of growing the list too big and/or setting limits which could be modifiable by the application if needed. 

I even think generating exceptions for out of size conditions would not be unreasonable.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
```diff
@@ -491,4 +574,32 @@ public static void DUMP_LOG_QUEUE() {
...
+  /**
+   * Gets a boolean indicating whether a specific log has been outputted. This is used to prevent
+   * spamming the log with the same message.
+   *
+   * @param logKey key for the log event
+   * @return {@code true} if the log has been outputted, {@code false} otherwise
+   * @since 1.15.9
+   */
+  public static boolean getLoggedOnce(String logKey) {
+    return logOnceList.contains(logKey);
```
it-hms, line 603: contains is O(n) for time
moving to a hash map ( or something similar) would be O(1). The idea is that the key would be the hash.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
it-hms: It's a good idea to cap the size to prevent resource issues, but we can add that later. It appears the aim of this PR is to address missing configuration options.

it-hms, line 94: I prefer hash map to array for faster lookups.

it-hms (changes requested):
One constraint this adds is that it requires unique config keys. I believe that two different JSON objects can have the same key but different meanings. So we just need to make this obvious.

alexjhawk:
Will update with `HashMap`, improved Javadocs, and uniqueness prefix in framework config `.logMissingField` methods.

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: I agree that this is a good idea, and is really applicable to many components in the library. I've created issue #86 to investigate and address this at a larger scale in the future.

alexjhawk: Migrated to a HashSet-based implementation, which in Java is backed by a HashMap.

src/main/java/com/hms_networks/americas/sc/extensions/connectors/framework/AbstractConnectorConfig.java
alexjhawk: Added a uniqueness prefix

src/main/java/com/hms_networks/americas/sc/extensions/logging/Logger.java
alexjhawk: Migrated to a HashSet-based implementation, which in Java is backed by a HashMap.

alexjhawk:
> Will update with `HashMap`, improved Javadocs, and uniqueness prefix in framework config `.logMissingField` methods.

- [X] Update with ~~`HashMap`~~`HashSet`
- [X] Improve Javadocs
- [X] Add framework config log key uniqueness prefix

it-hms (approved):

TomKimsey (approved):

PR #99, opened 2024-08-01
Release/1.16.0 pre2

it-hms:
This PR is a pre-release which will be followed up by an additional PR.
Changes:  
- support for tag instant value EBD calls 
- DataPoint classes now support ISO 8601 timestamp 
- Datapoint epoch second timestamps are depreciated, but still included for compatibility 
- Historical Data EBD calls are aware of store data in UTC setting and ISO 8601 timestamps will reflect that setting

Because epoch seconds timestamps support remains, but is deprecated, there is no major revision change planed.

### DIFF
--- a/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
@@ -19,6 +19,9 @@ class HistoricalDataConstants {
   /** Index of time int in EBD lines */
   public static final int EBD_LINE_TAG_TIMEINT_INDEX = 1;
 
+  /** Index of time string in EBD lines */
+  public static final int EBD_LINE_TAG_TIMESTR_INDEX = 2;
+
   /** Index of tag value in EBD lines */
   public static final int EBD_LINE_TAG_VALUE_INDEX = 4;
 
@@ -145,4 +148,13 @@ class HistoricalDataConstants {
 
   /** The IO server used for queue diagnostic tag(s). */
   public static final String QUEUE_DIAGNOSTIC_TAG_IO_SERVER = "MEM";
+
+  /** Start index of the actual ISO 8601 timestamp return by EBD. */
+  public static final int EBD_ISO8601_TIMESTAMP_START_INDEX = 1;
+
+  /** String index where milliseconds should be added to a timestamp that lacks milliseconds. */
+  public static final int EBD_ISO8601_TIMESTAMP_START_MILLIS = 20;
+
+  /** String to add to a timestamp that lacks milliseconds. */
+  public static final String EBD_ISO8601_TIMESTAMP_ADD_MILLIS_STR = ".000";
 }
--- a/src/main/java/com/hms_networks/americas/sc/extensions/realtimedata/InstantValuesEbdString.java
+++ b/src/main/java/com/hms_networks/americas/sc/extensions/realtimedata/InstantValuesEbdString.java
@@ -1,0 +1,581 @@
+package com.hms_networks.americas.sc.extensions.realtimedata;
+
+import com.ewon.ewonitf.Exporter;
+import com.hms_networks.americas.sc.extensions.datapoint.DataPoint;
+import com.hms_networks.americas.sc.extensions.datapoint.DataPointBoolean;
+import com.hms_networks.americas.sc.extensions.datapoint.DataPointNumber;
+import com.hms_networks.americas.sc.extensions.datapoint.DataPointString;
+import com.hms_networks.americas.sc.extensions.datapoint.DataQuality;
+import com.hms_networks.americas.sc.extensions.historicaldata.HistoricalDataManager;
+import com.hms_networks.americas.sc.extensions.string.StringUtils;
+import com.hms_networks.americas.sc.extensions.system.time.SCTimeUnit;
+import com.hms_networks.americas.sc.extensions.system.time.SCTimeUtils;
+import com.hms_networks.americas.sc.extensions.taginfo.TagInfo;
+import com.hms_networks.americas.sc.extensions.taginfo.TagInfoManager;
+import com.hms_networks.americas.sc.extensions.taginfo.TagType;
+import com.hms_networks.americas.sc.extensions.util.StreamUtils;
+import java.util.ArrayList;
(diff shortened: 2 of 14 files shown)


src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
```diff
@@ -145,4 +148,13 @@ class HistoricalDataConstants {
...
   public static final String QUEUE_DIAGNOSTIC_TAG_IO_SERVER = "MEM";
+
+  /** Start index of the actual ISO 8601 timestamp return by EBD. */
+  public static final int EBD_ISO8601_TIMESTAMP_START_INDEX = 1;
+
+  /** String index where milliseconds should be added to a timestamp that lacks milliseconds. */
+  public static final int EBD_ISO8601_TIMESTAMP_START_MILLIS = 20;
+
+  /** String to add to a timestamp that lacks milliseconds. */
+  public static final String EBD_ISO8601_TIMESTAMP_ADD_MILLIS_STR = ".000";
```
alexjhawk, line 159: This is more of a question than an issue, but was there an issue parsing without manually adding the `.000`?

alexjhawk (approved):
I didn't notice anything in the PR, although I may be a bit biased on the overall/structural aspects, having written some of the code.

it-hms:
> I didn't notice anything in the PR, although I may be a bit biased on the overall/structural aspects, having written some of the code.

One thing to review/chime in on is the DataPoint changes to have two timestamps. This is to preserve legacy applications and update to better handling of timezone offsets

src/main/java/com/hms_networks/americas/sc/extensions/historicaldata/HistoricalDataConstants.java
it-hms: This is for parsing but also because most connectors ( this true for Azure - not sure about the others ) have ISO 6801 timestamps with millisecond resolution. I didn't want to introduce seconds and have the connectors add milliseconds for compatibility.   
Actually, I made a seconds timestamp and a milliseconds timestamps (and two getter methods) and then deleted the seconds because I thought it was unnecessary bloat.

alexjhawk:
> > I didn't notice anything in the PR, although I may be a bit biased on the overall/structural aspects, having written some of the code.
> 
> One thing to review/chime in on is the DataPoint changes to have two timestamps. This is to preserve legacy applications and update to better handling of timezone offsets

This is definitely good. I recall *attempting* to switch to one timestamp, too, but ended up running into issues with both the ISO-8601 timestamp and numerical timestamp being required in certain places.

pom.xml
```diff
@@ -8,7 +8,7 @@
   <!-- PROJECT NAME -->
   <name>Ewon Flexy Extensions Library</name>
   <!-- PROJECT VERSION -->
-  <version>1.15.14</version>
+  <version>1.16.0-pre2</version>
```
TomKimsey, line 11: @it-hms Why isn't this 1.16.0?

web-docs/docs/02-CHANGELOG.mdx
```diff
@@ -5,6 +5,10 @@ sidebar_label: Change Log
 toc_max_heading_level: 2
 ---
 
+## Version 1.16.0-pre2
+### Features
+- Lorem ipsum
+
```
TomKimsey, line 11: Fill this in

pom.xml
TomKimsey, line 11: We haven't been committing pre-releases to main. If this is not ready to be a full release this commit can be dropped.

src/main/java/com/hms_networks/americas/sc/extensions/realtimedata/InstantValuesEbdString.java
```diff
@@ -0,0 +1,581 @@
...
+ * 2003;"Str_100";"ABCDEF\"GH\'IJKLMNOPQRSTUVWXYZ";0;0;65472
+ * 2004;"Str_101";"ABCDEFGHIJKLMNOPQRSTUVWXYZ";0;0;65472
+ * 2005;"Str_102";"AT char @";0;0;65472
+ * </pre>
+ *
+ * @author HMS Networks; Americas
+ * @version 1.0.0
+ * @since 1.15.15
+ */
+public class InstantValuesEbdString {
```
TomKimsey, line 65: Have you checked the performance of this vs the TagControl class? EBDs are very costly and it might be more performant to use TagControl even for large amounts of tags.

TomKimsey (changes requested):

web-docs/docs/02-CHANGELOG.mdx
TomKimsey: Not relevant if this commit gets dropped

pom.xml
TomKimsey: If you are doing this for maven reasons i think Alex generally just pushes a separate branch and tags it.

it-hms: It's common for repositories to have the main branch represent the bleeding edge of development and recommend a release branch/tag for releases.  Updating the version tag in the pom is useful because `mvn deploy` will update the local maven repository and then local connector development/testing can use this development version.

src/main/java/com/hms_networks/americas/sc/extensions/realtimedata/InstantValuesEbdString.java
it-hms: Instant value EBD is roughly1100ms.  TagControl is 10ms per tag. EBD instant returns all tags. TagControl is each tag. The break even is around 110 tags. 
I think we can live with this performance, and improve the CPU utilization by implementing a non-CPU blocking wait for EBD response.

pom.xml
it-hms: will drop this commit

it-hms:
Dropped the last pre-release commit

pom.xml
alexjhawk: Our main branch should represent the bleeding edge of development (in most cases), though we have just historically excluded pre-release tagging commits. The code/content commits themselves definitely belong in the main branch, but any tagged pre-release commits are typically left out and pushed+tagged directly to save time.

alexjhawk (approved):

TomKimsey (approved):

